PR #4, opened 2023-05-16
Support iOS 12.0

KasparElmans:
From XCode 14.3 and on, the minimum deployment target for a framework and its dependencies need to be 12.0 or higher. The current Maplibre Navigation SDK uses a couple of dependencies that still use 8.0 as a minimum. In this PR we have updated the dependencies accordingly and fixes minor issues. 

As [discussed](https://github.com/maplibre/maplibre/issues/278), we will host the upgraded MapboxDirections SDK at Flitsmeister until this SDK is properly updated. 

While updating Turf to the latest version or at least v1.0.0 we encountered some breaking unit test in de Maplibre Navigation SDK. It appears Turf has some bug fixes, which also has effect on this SDK. It could either be wrong Unit Tests or a wrong implementation. We are unable to find what changes in Turf have caused this issue. Since we can't be certain about the cause we decided to fork the currently used Turf as well and update it to 12.0. We will host this fork until it is properly updated and the unit tests have been fixed.


louwers (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 - Moved & renamed `RouteControllerProactiveReroutingInterval` to be an instance property of `RouteController` so the client can easily set this per route controller
 - Added `shouldReturnTestingETAUpdateReroutes` property to `RouteController` as an easy way to test an ETA update client-side. It uses two of the same test-routes between `52.02224357,5.78149084` and `52.03924958,5.55054131` with different ETA's to easily see the ETA change happen in the client's UI
 
+## v1.0.7 (May 23, 2023)
+ - Upgrade iOS version from 11.0 to 12.0.
+ - Upgraded dependencies to support iOS 12.0
+
```
Svantulden, line 14: There's also a v1.0.7 above this in the changelog, is the versioning correct?

Svantulden: Also might want to specifically specify the upgrade to be the _minimum_ iOS version for clarity, as the SDK already supported iOS 12

Svantulden (approved):
Some small nitpickings, but good that we now have a version that works with XCode 14.3 👍

CHANGELOG.md
KasparElmans: I have updated the changelog accordingly.

Cartfile
```diff
@@ -1,6 +1,5 @@
...
-github "mapbox/mapbox-events-ios" "v0.10.2"
-github "mapbox/MapboxDirections.swift" ~> 0.23.0
-github "mapbox/turf-swift" ~> 0.2
-github "ceeK/Solar" ~> 2.1.0
-github "mapbox/mapbox-voice-swift" ~> 0.0.1
+github "mapbox/MapboxGeocoder.swift" ~> 0.15
+github "flitsmeister/Mapbox-directions-swift" "v0.23.1"
+github "flitsmeister/turf-swift" "v0.2.2"
+github "ceeK/Solar" ~> 3.0.1
+github "mapbox/mapbox-speech-swift" "v2.1.1"
```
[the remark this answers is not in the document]
KasparElmans: Yes it does -> https://github.com/mapbox/mapbox-speech-swift/blob/main/LICENSE.md

Diego5529 (approved):

louwers:
@Svantulden @KasparElmans I see the changelog is updated.

However, this is a breaking release so it should be released as 2.0.0 instead right?

@birkskyum made a release last time. Any volunteers for releasing it this time?

KasparElmans:
> @Svantulden @KasparElmans I see the changelog is updated.
> 
> However, this is a breaking release so it should be released as 2.0.0 instead right?
> 
> @birkskyum made a release last time. Any volunteers for releasing it this time?

louwers:
@KasparElmans Looks like it!

louwers:
@KasparElmans Let's try! :)

PR #8, opened 2023-10-19
Smooth camera transition when following user

### LINKED ISSUE #1: Camera does not follow the user properly in NavigationViewController
I am trying to get this project working in a non-Mapbox environment and for some reason the camera is not properly following the user. My understanding is that this is supposed to happen automatically.

To elaborate a bit on what non-Mapbox looks like, we are using OpenMapTiles-derived tiles (Stadia Maps) in place of Mapbox map tiles, and are constructing the `Route` directly from an OSRM JSON response (in this case, generated by the Stadia Maps Valhalla API; to save the trouble for those not familiar, the Mapbox Directions API v5 only adds a few minor things, mostly related to voice navigation, to the OSRM response format, and these are optional).

### Steps to Trigger Behavior

1. Create a `Route` (in this case, parsed from an OSRM response)
2. Create a `Style` pointing to a Stadia Maps URL (I suspect any non-Mapbox URL will do; ex: MapTiler if you prefer)
3. Create a `SimulatedLocationManager` with the route and set a speed multiplier of 5 so you aren't waiting all day ;)
4. Create a `NavigationViewController` using the route, style and simulated location manager from above
5. Present the view controller

https://github.com/ianthetechie/maplibre-ios-camera-mre. Two notes on this. 1) you will need a (free) Stadia Maps API key, which obviously isn't published (edit Config.swift). 2) I'm referencing a fork of maplibre-navigation-ios to remove a few mapbox dependencies, but the issue appears with the official release as well.

### Expected Behavior

The user "puck" should move along the route line, the camera heading/bearing should adjust to the user, and he camera should move "forward" along with the user puck.

### Actual Behavior

The user puck moves along the route line. While the camera DOES change in some manner (bearing/heading only), it never actually follows the user, or even fully changes bearing to point toward the user.

https://user-images.githubusercontent.com/2140440/218034990-562ece68-373f-4540-a513-1e4405de1b26.MP4

I cut the video off a bit early but it never catches up to the user or pans the correct direction.

### Version(s) affected

- Platform: iOS
- OS version: 16.3
- Device type: iPhone 13 Mini
- Emulator/ Simulator: no
- Development OS: macOS Ventura 13.1
- maplibre-navigation-ios Version 1.0.7
- MapLibre GL version 5.12.2

### Additional context

I forked the repo because the current road info was not showing and I was getting console errors about the `mapbox://` URL scheme not being supported. Is it possible that the dependency on the street styles is responsible in some way? I'm poking around `RouteMapViewController` as this seems to be a likely spot for camera manipulation but it's going to take me a bit to parse it all.

BroxZaphod:
Fixes #1 

The func `layoutSubviews` is called each time the mapView changes, causing `updateCourseTracking`
 called in a too high frequency. Also setting the `edgePadding` in `setCamera()` to `self.cameraEdgeInsets` instead of adding to high padding values will remove the jerky map transitions.

By setting a custom camera in the Example App you can add custom `UIEdgeInsets`.
Even without the custom camera the Example App will work smoothly.

- Removed updateCourseTracking from layoutSubviews()
- Set self.cameraEdgeInsets for each setCamera in updateCourseTracking()

UPDATE:

Removing the camera updates from `updateCourseTracking` reduces the call about 30-40 times a second. The example App should run now without the jerky map animations. Also works fine when setting a custom camera.

- Added layoutSubviews() again
- Removed `setCamera()` from `updateCourseTracking()`
- Added `setCamera()` to `progressDidChange()`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Changes to the Mapbox Navigation SDK for iOS
 
 ## Unreleased
- - Removed `updateCourseTracking` from `layoutSubviews()`
- - Set `self.cameraEdgeInsets` for each `setCamera` in `updateCourseTracking()`
+ - Removed setCamera() from updateCourseTracking()
+ - Added setCamera() to progressDidChange()
 
 ## v2.0.0 (May 23, 2023)
  - Upgrade minimum iOS version from 11.0 to 12.0.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changes to the Mapbox Navigation SDK for iOS
 
+## Unreleased
+ - Removed `updateCourseTracking` from `layoutSubviews()`
+ - Set `self.cameraEdgeInsets` for each `setCamera` in `updateCourseTracking()`
+
 ## v2.0.0 (May 23, 2023)
  - Upgrade minimum iOS version from 11.0 to 12.0.
  - Upgraded dependencies to support iOS 12.0
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -244,15 +244,6 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         addSubview(imageView)
     }
     
-    open override func layoutSubviews() {
-        super.layoutSubviews()
-        
-        //If the map is in tracking mode, make sure we update the camera after the layout pass.
-        if (tracksUserCourse) {
-            updateCourseTracking(location: userLocationForCourseTracking, camera:self.camera, animated: false)
-        }
-    }
-    
     open override func anchorPoint(forGesture gesture: UIGestureRecognizer) -> CGPoint {
         if tracksUserCourse {
             return userAnchorPoint
@@ -342,11 +333,9 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         }
         
         if tracksUserCourse {
-            let point = userAnchorPoint
-            let padding = UIEdgeInsets(top: point.y, left: point.x, bottom: bounds.height - point.y, right: bounds.width - point.x)
-            let newCamera = camera ?? MGLMapCamera(lookingAtCenter: location.coordinate, fromDistance: altitude, pitch: 45, heading: location.course)
+            let newCamera = camera ?? MGLMapCamera(lookingAtCenter: location.coordinate, altitude: altitude, pitch: 45.0, heading: location.course)
             let function: CAMediaTimingFunction? = animated ? CAMediaTimingFunction(name: CAMediaTimingFunctionName.linear) : nil
-            setCamera(newCamera, withDuration: duration, animationTimingFunction: function, edgePadding: padding, completionHandler: nil)
+            setCamera(newCamera, withDuration: duration, animationTimingFunction: function, edgePadding: self.cameraEdgeInsets, completionHandler: nil)
         }
         if !tracksUserCourse || userAnchorPoint != userCourseView?.center ?? userAnchorPoint {
             UIView.animate(withDuration: duration, delay: 0, options: [.curveLinear, .beginFromCurrentState], animations: {
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -244,6 +244,15 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         addSubview(imageView)
     }
     
+    open override func layoutSubviews() {
+        super.layoutSubviews()
+        
+        //If the map is in tracking mode, make sure we update the camera after the layout pass.
+        if (tracksUserCourse) {
+            updateCourseTracking(location: userLocationForCourseTracking, camera:self.camera, animated: false)
+        }
+    }
+    
     open override func anchorPoint(forGesture gesture: UIGestureRecognizer) -> CGPoint {
         if tracksUserCourse {
             return userAnchorPoint
@@ -335,7 +344,7 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
         if tracksUserCourse {
             let newCamera = camera ?? MGLMapCamera(lookingAtCenter: location.coordinate, altitude: altitude, pitch: 45.0, heading: location.course)
             let function: CAMediaTimingFunction? = animated ? CAMediaTimingFunction(name: CAMediaTimingFunctionName.linear) : nil
-            setCamera(newCamera, withDuration: duration, animationTimingFunction: function, edgePadding: self.cameraEdgeInsets, completionHandler: nil)
+            setCamera(newCamera, withDuration: duration, animationTimingFunction: function, edgePadding: UIEdgeInsets.zero, completionHandler: nil)
         }
         if !tracksUserCourse || userAnchorPoint != userCourseView?.center ?? userAnchorPoint {
             UIView.animate(withDuration: duration, delay: 0, options: [.curveLinear, .beginFromCurrentState], animations: {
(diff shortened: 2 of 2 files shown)


wipfli:
@boltrn since you are so familiar with the android version of the navigation plugin, could you maybe review this pull request?

boldtrn (approved):
Thanks for providing this PR, the changes look good to me.

mindthefish:
Previously I tried the same (not calling the `updateCourseTracking`  on `layoutSubviews` calls). Unfortunately I ended up with strange free-floating puck behavior near maneuvers. But this might also be related to my custom routing backend. I will try again with your changes.

mindthefish:
I am sorry to say that removing `updateCourseTracking` from `layoutSubviews` causes problems for me. Attached are two videos, the first with the `updateCourseTracking` call in `layoutSubviews` (state on main), the second video without (pull request). Can anyone confirm or disconfirm this?
I can't say anything about the `self.cameraEdgeInsets` yet, because as a workaround I currently just set the padding to `UIEdgeInsets.zero` and control my puck position via `contentInsets`.


https://github.com/maplibre/maplibre-navigation-ios/assets/895046/44bd98a9-9754-4c28-bc87-0c14963962c1

https://github.com/maplibre/maplibre-navigation-ios/assets/895046/0cd3a1b5-c37f-45c7-9db4-f067cf8e1329

boldtrn:
Thanks for having a look at this PR @mindthefish. It's interesting that you are seeing this. 

Just to clarify, could this be related to your custom implementation? Routing-backend or custom-ui? Can you reproduce the same result with the regular ui and with a route response from Mapbox? 

Both the puck and the camera are animated and should be called about once per second. If you look at `layoutSubviews`, this is triggered several times per second and is updating the camera too frequentl.y.

Dennis-Iatropoulos:
Hey @mindthefish, thanks for testing!

In my custom implementation i can´t see the behaviour seen in your video. So i used the `Example-Swift` target provided in the Framework.
I added the `layoutSubviews` again and add the `UIEdgeInsets.zero` directly. This should work without setting a own camera in a custom implementation.
- In the first video you see the current `main` branch. (Laggy animations)
- The second one shows the smoother map-animations in usertracking-mode

Hope this will work when you build the Example App and your App.
If yes, we have a working Example App in the Framework.

https://github.com/maplibre/maplibre-navigation-ios/assets/51818940/a5109e5a-8261-4f8f-8333-55c25f9f3b69

https://github.com/maplibre/maplibre-navigation-ios/assets/51818940/5df53f7a-4de9-44ce-a975-190e1568a371

mindthefish:
> Hey @mindthefish, thanks for testing!
> 
> In my custom implementation i can´t see the behaviour seen in your video. So i used the `Example-Swift` target provided in the Framework. I added the `layoutSubviews` again and add the `UIEdgeInsets.zero` directly. This should work without setting a own camera in a custom implementation.
> 
> * In the first video you see the current `main` branch. (Laggy animations)
> * The second one shows the smoother map-animations in usertracking-mode
> 
> Hope this will work when you build the Example App and your App. If yes, we have a working Example App in the Framework.
> 

I see this more as a MapLibre Native related issue. The edgePadding should be used to set the camera offset and with it the puck position on the screen. Setting the camera too often with a non-zero edgePadding seems to cause an issue with the camera animation. Using no padding (UIEdgeInsets.zero) solves the camera animation but is more like a workaround. This issue has been addressed at https://github.com/maplibre/maplibre-native/issues/958 but there has not been any progress yet.
I will try to run the example app with the Mapbox Directions API tomorrow.

Dennis-Iatropoulos:
Thanks @mindthefish for pointing that out.
This should be fixed in `maplibre-native`. Better than a workaround here.

mindthefish:
I ran the example app with a Mapbox Directions route (sorry, I did not get the default map style to work) and although it does not make sense to call `updateCourseTracking` something like 40x per second from `layoutSubviews`, I still have the impression, that the puck sticks better to the route by leaving it like it is. Can somebody confirm this?

First video: without excessive `updateCourseTracking`, second video with:


https://github.com/maplibre/maplibre-navigation-ios/assets/895046/41cf3756-b16b-4d9c-b3fb-7b1a7bec1e33

https://github.com/maplibre/maplibre-navigation-ios/assets/895046/e64f8d29-1896-4b63-888e-d3c40cf9ca41

mindthefish:
> > I also tried to build maplibre-native for iOS from source to look into this issue but did not get it to work until now.
> 
> I would recommend to let them know, they recently switched to Bazel which changes how you build it. @louwers - FYI. You might want to join the OSMUS slack, there have been some discussions about this.

Already there, need to have a look more often. Thanks for the hint!

torrswanson:
I've removed `updateCourseTracking` in `layoutSubviews` when not using `.zero` for `edgePadding` in the `setCamera(..)`.

I'm using `let padding = UIEdgeInsets(top: point.y, left: 0.0, bottom: bounds.height - point.y, right: 0.0)`. Specifically setting 0 values for `left` and `right`. This means not using the `self.cameraEdgeInsets` that is propsed in this pull request and leaving the padding value based off the `userAnchorPoint` that itself can be based off a delegate method. I'd rather leave `top` and `bottom` values as-is. There should be a really really good reason to get rid of using `userAnchorPoint` for this PR to be approved as the top and bottom values at runtime are different as shown below.

On my iPhone 8, these are the values i'm getting at runtime:
`self.cameraEdgeInsets` is `UIEdgeInsets(top: 597.6, left: 0.0, bottom: 125.39999999999998, right: 0.0)`
`padding` is `UIEdgeInsets(top: 501.6, left: 0.0, bottom: 125.39999999999998, right: 0.0)`

I'm not using `contentInsets` on the NavigationMapView because it alters the Attribution marks.

boldtrn:
@torrswanson thanks for chiming in, I agree we definitely need insets/padding for the camera. Would like to propose a different approach and would like to provide a PR for this?

Diego5529 (approved):

mindthefish:
Did we agree that this PR should not be merged?

boldtrn:
In general I would be fine with merging this PR, but it seems a bit like a workaround and not like an actual fix? @torrswanson mentioned a different approach, I am not sure if it's worth to check this approach?

boldtrn:
I did have a look at the PR, the results are way better now with the latest changes, we don't call the camera multiple times per second anymore. 

We still need to update the README.

In a follow-up we should look into extendability so that implementers can actually change the way the camera works.

mindthefish:
Sorry to bother you again about this PR. Since the merge, I've been experiencing weird behaviour in the puck animation. This behaviour occurs during a manoeuvre and is, in my opinion, due to the fact that the camera update now happens less frequently and, in combination with the snapping to the new segment, leads to an awkward animation.

This behaviour does not occur for me with the SimulatedLocationManager, as this distributes the artificial positions perfectly along the route. I record real GPS positions for debugging (in this case a motorbike tour) so that I can work with the same data later.

The following image shows the individual location updates, each point is one location update. You can see that several updates are very close to each other before the turning maneuver while waiting for oncoming traffic, but there are large gaps during the turning maneuver itself. I think this is a realistic scenario that occurs frequently.
IMG_0460

The first video shows the behaviour before the pull request, the second video shows the behaviour after the pull request.

https://github.com/maplibre/maplibre-navigation-ios/assets/895046/86b07033-57b7-4994-be64-2967d451bc55

https://github.com/maplibre/maplibre-navigation-ios/assets/895046/90176c8e-fbae-4fd1-b96b-905fc405c7b5

I'm not saying that the PR is faulty, but in conjunction with snapping to the route and aligning the puck, it causes problems.

boldtrn:
@mindthefish since this PR is merged, maybe we should continue the discussion in a new issue? 

Generally: I agree and we should avoid updating the puck more frequently than the camera. I believe we should minimize the puck updates to every location update and add an animation duration to the puck update. I believe @BroxZaphod  saw issues with this approach, so it's probably not super straight forward. 

I would be happy to check out any PRs looking into this. So if you would like to help improve this, that would be great :).

PR #11, opened 2023-12-26
Allow to overwrite update camera via delegate

boldtrn:
Make it possible to customize the camera update through a delegate for implementing apps.


MapboxNavigation/NavigationMapView.swift
```diff
@@ -1244,4 +1249,13 @@ public protocol NavigationMapViewCourseTrackingDelegate: AnyObject {
...
     optional func navigationMapViewDidStopTrackingCourse(_ mapView: NavigationMapView)
+    
+    /**
+     Allows to pass a custom camera location given the current route progress. If you return true, the camera won't be updated by the NavigationMapView.
+     - parameter mapView: The NavigationMapView.
+     - parameter location: The current user location
+     - parameter routeProgress: The current route progress
+     */
+    @objc(navigationMapView:location:routeProgrss:)
+    optional func updateCamera(_ mapView: NavigationMapView, location: CLLocation, routeProgress: RouteProgress) -> Bool
```
ianthetechie, line 1260: Just one minor nitpick on the way that delegate APIs are usually designed on Apple platforms. See [UITextViewDelegate](https://developer.apple.com/documentation/uikit/uitextviewdelegate/1618630-textview) for an example of how these are usually named. The logic is also usually inverted so that if the method returns true (or is unimplemented), the user accepts the defaults.

```swift
 @objc(navigationMapView:shouldUpdateCameraTo:userLocation:routeProgress:)
    optional func navigationMapView(_ mapView: NavigationMapView, shouldUpdateCameraTo: MGLCamera, userLocation: CLLocation, routeProgress: RouteProgress) -> Bool
```

boldtrn: Thanks for looking into this :)

The delegates are already somewhat inconsistently renamed in this project. Some methods use the naming you propose. 

For readability I prefer properly named methods instead of tons of the same methods names with different parameter names. Do you think this will be irritating for iOS devs?

If I look into UITextViewDelegate, there are also methods with different naming, [for example](https://developer.apple.com/documentation/uikit/uitextviewdelegate/1618608-textviewshouldbeginediting).

If this is something else, please let me know, it would be great to learn more about Swift 👍

ianthetechie: > Thanks for looking into this :)

Sure thing! I'll get to the other one about rerouting logic at some point this week as well.

> The delegates are already somewhat inconsistently renamed in this project. Some methods use the naming you propose.
>
> For readability I prefer properly named methods instead of tons of the same methods names with different parameter names. Do you think this will be irritating for iOS devs?

Yeah, I gathered that 😂 In which case, as long as it's _reasonably_ consistent, I guess it's at least not _new_  entropy. I would however still fix the typo in the objc signature (`Progrss`).

> If I look into UITextViewDelegate, there are also methods with different naming, [for example](https://developer.apple.com/documentation/uikit/uitextviewdelegate/1618608-textviewshouldbeginediting).

This is actually following the same convention. The general pattern is something like `<object>should<Action>` for cases where there is only a single parameter, and `<object>:should<Action><preposition or something>:<otherParameters>`. The conventions are admittedly not intuitive at first, and is very much "legacy" going back to the NEXT days and obj-c, but Apple did codify some of it a few years ago in the [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/). Not the easiest reading but if you're bored and want to get deeper into Swift it might be helpful to read at some point (I think it's most helpful to just write a bunch of code, using a bunch of Apple's APIs in practice, and then read the guidelines later as it's hard to just sit down and read, but is a lot easier to skim to solidify the patterns that you'd pick up on subconsciously).

boldtrn: Thanks a lot for the input. I will try to improve my Swift skills :). I have updated the typo and will go ahead and merge this PR.

PR #12, opened 2023-12-27
Remove MapboxSpeechController and Mapbox Speech dependency

boldtrn:
Given that we try to get rid of Mapbox dependencies here, I created a PR that removes the MapboxSpeech dependency and implementation.

If you would like to continue Mapbox Speech, you can copy the deleted MapboxSpeechController into your project.

IMHO, we should not include any Mapbox dependencies (and should remove the other dependencies here in the code as well).


louwers:
Sounds reasonable, but I am not really involved with the navigation repos at this point. :)

boldtrn:
@louwers Got it. Sorry if I tagged the wrong people here, I wasn't sure who would be the right person to request reviews from?

louwers:
I know @ianthetechie is into using MapLibre for navigation, but his efforts are probably focused on https://github.com/stadiamaps/ferrostar

ianthetechie:
Hey, thanks fro the mention @louwers :)

Yes, my efforts are indeed mostly focused on Ferrostar, but I'm also happy to help out here if I can (though my time is understandably limited). I really do want the MapLibre Navigation projects to succeed as well; I'm exploring new approaches because I think it'll be easier to sustain something longer-term that's vendor-neutral first rather than a Mapbox fork in the case of the nav SDKs :D

For this specific PR, I think the changes are good. It looks like it'll still maintain speech synthesis support via the Apple local APIs, which are exactly what I think most people using this library (or Ferrostar) will want anyways.

boldtrn:
Thanks for the feedback @ianthetechie. 

I agree and I really appreciate the efforts you are doing with Ferrostar. If we had the time and budget to fully rebuild a new nav sdk, then this would be my favorite option as well. For the time being, keeping the nav sdks working well is IMHO the best option.

If you have the time it, I'd appreciate feedback on the other PRs as well. But I can totally understand the limited time issue, I am struggling with finding enough time as well :).

> For this specific PR, I think the changes are good. It looks like it'll still maintain speech synthesis support via the Apple local APIs, which are exactly what I think most people using this library (or Ferrostar) will want anyways.

Yes, that's what I was thinking as well. If anyone wants to implement MapboxVoice (or any other server side TTS), that's still possible but is out of scope of this repo.

PR #16, opened 2023-12-31
Overwrite rerouting logic

boldtrn:
This changes makes it possible to write custom rerouting logic in your app without requiring to use Mapbox directions.


MapboxCoreNavigation/RouteControllerDelegate.swift
```diff
@@ -114,4 +114,17 @@ public protocol RouteControllerDelegate: AnyObject {
      */
     @objc(routeControllerShouldDisableBatteryMonitoring:)
     optional func routeControllerShouldDisableBatteryMonitoring(_ routeController: RouteController) -> Bool
+    
+    /**
+    Allows to customise the calculation of a route.
+    
+    If you want to overwrite the defult rerouting loigc, return true.
```
mindthefish, line 121:
```suggestion
    If you want to overwrite the default rerouting logic, return true.
```

MapboxCoreNavigation/RouteControllerDelegate.swift
```diff
@@ -114,4 +114,17 @@ public protocol RouteControllerDelegate: AnyObject {
...
+     
+     - parameter from: The current olocation of the user
```
mindthefish, line 123:
```suggestion
     - parameter from: The current location of the user
```

MapboxCoreNavigation/RouteControllerDelegate.swift
```diff
@@ -114,4 +114,17 @@ public protocol RouteControllerDelegate: AnyObject {
...
+     - parameter along: The route progress
+     - parameter completion: Callbackc function when either the route was sucessfully caluclated or if there was an error
```
mindthefish, line 125:
```suggestion
     - parameter completion: Callback function when either the route was successfully calculated or if there was an error
```

MapboxCoreNavigation/RouteControllerDelegate.swift
```diff
@@ -114,4 +114,17 @@ public protocol RouteControllerDelegate: AnyObject {
...
+     - return: True to preferent the route controller from running its own rerouting logic
```
mindthefish, line 126:
```suggestion
     - return: True to prevent the route controller from running its own rerouting logic
```

mindthefish (changes requested):
Just some typos

boldtrn:
Thanks, I just found the spellchecker in XCode, I will try my best to improve the spelling :)

ianthetechie (approved):
Sorry for the delay in reviewing this, but this looks good! Can't wait to try it out! Thanks @boldtrn!

boldtrn:
Thanks for the review @ianthetechie, I will go ahead and merge this PR.

PR #19, opened 2024-01-07
Only snap location if user is close to the route line

boldtrn:
This PR will change the snapping logic. Right now the location will be snapped on the route line, no matter how far a user is away from the route line, as long as the user is facing into the somewhat right direction.

This PR make sure the snapping only happens if the user is close to the route line and faces in the right directions.

This PR still needs a bit more real life testing, first quick tests look good.

@ianthetechie if you have the time, a review would be very welcome.

### DIFF
--- a/MapboxCoreNavigation/CLLocation.swift
+++ b/MapboxCoreNavigation/CLLocation.swift
@@ -74,14 +74,18 @@ extension CLLocation {
     func snapped(to legProgress: RouteLegProgress) -> CLLocation? {
         let coords = coordinates(for: legProgress)
         
-        guard let closest = Polyline(coords).closestCoordinate(to: coordinate) else { return nil }
+        guard let closest = Polyline(coords).closestCoordinate(to: self.coordinate) else { return nil }
         guard let calculatedCourseForLocationOnStep = interpolatedCourse(along: coords) else { return nil }
         
         let userCourse = calculatedCourseForLocationOnStep
         let userCoordinate = closest.coordinate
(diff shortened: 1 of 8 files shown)


MapboxCoreNavigation/CLLocation.swift
```diff
@@ -159,16 +163,16 @@ extension CLLocation {
     }
     
     /**
-     Determines if the a location is qualified enough to allow the user puck to become unsnapped.
+     Determines if the course of a location is qualified enough to allow the user puck to be snapped
      */
-    func shouldSnap(toRouteWith course: CLLocationDirection, distanceToFirstCoordinateOnLeg: CLLocationDistance = CLLocationDistanceMax) -> Bool {
+    func shouldSnapCourse(toRouteWith course: CLLocationDirection, distanceToFirstCoordinateOnLeg: CLLocationDistance = CLLocationDistanceMax) -> Bool {
```
mindthefish, line 168: Is there a reason for renaming the function?

boldtrn: Yes, because this function is only checking the course. The name was misleading before. We could also do the distance check in this function, but this would require to calculate it again (as this is done in a different piece of the code).

boldtrn:
Thanks for looking into this, I actually wanted to merge this quite some time ago already, we are using this version and it's working pretty well so far.I am happy to hold off with merging until SPM & Maplibre Update is done, but we could merge this sooner as well.

mindthefish (approved):

PR #23, opened 2024-02-02
chore: updated MapLibre Native dependency to use the ios-v6.0.0 versi…

mindthefish:
updated MapLibre Native dependency to use the ios-v6.0.0 version (https://github.com/maplibre/maplibre-native/releases/tag/ios-v6.0.0).
Full refactoring to MapLibre / MLN naming convention still to be done.
Please test and review in-depth, as most of the refactoring was made on a find&replace base.

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -1,23 +1,23 @@
 import Foundation
-import Mapbox
+import MapLibre
 import MapboxDirections
 import MapboxCoreNavigation
 import Turf
 
 /**
- `NavigationMapView` is a subclass of `MGLMapView` with convenience functions for adding `Route` lines to a map.
+ `NavigationMapView` is a subclass of `MLNMapView` with convenience functions for adding `Route` lines to a map.
  */
-@objc(MBNavigationMapView)
-open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
+@objc(MLNavigationMapView)
+open class NavigationMapView: MLNMapView, UIGestureRecognizerDelegate {
     
     // MARK: Class Constants
     
     struct FrameIntervalOptions {
         fileprivate static let durationUntilNextManeuver: TimeInterval = 7
         fileprivate static let durationSincePreviousManeuver: TimeInterval = 3
-        fileprivate static let defaultFramesPerSecond = MGLMapViewPreferredFramesPerSecond.maximum
-        fileprivate static let pluggedInFramesPerSecond = MGLMapViewPreferredFramesPerSecond.lowPower
-        fileprivate static let decreasedFramesPerSecond = MGLMapViewPreferredFramesPerSecond(rawValue: 5)
+        fileprivate static let defaultFramesPerSecond = MLNMapViewPreferredFramesPerSecond.maximum
+        fileprivate static let pluggedInFramesPerSecond = MLNMapViewPreferredFramesPerSecond.lowPower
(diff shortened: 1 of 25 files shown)


boldtrn:
Thank you very much for looking into this. I will do some testing.

What's your general feeling testing this, should we rather try to merge this ASAP or keep open for testing for some time?

mindthefish:
For me, running the sample application works as it did before the refactoring. However, I had problems loading the Mapbox map style before and after, despite the access key (or does the style only consist of single-colored country areas without roads?). So setting the pin, retrieving the route, starting active navigation, voice output, worked without problems in my tests. 
I would prefer to merge as soon as possible, especially to make quick progress with the SPM PR that has just been closed. Still, I'm relatively new to Swift, so I'd be grateful if someone with more expertise would take a few minutes to review. Thank you!

boldtrn:
> or does the style only consist of single-colored country areas without roads?

Yes, this is currently intended, you can see this style used a few times here: https://maplibre.org/

You can set your own map style to overwrite this.

> I would prefer to merge as soon as possible

I think this would be the best approach as well 👍 

> Still, I'm relatively new to Swift

Yep, same here :)

boldtrn:
BTW: the CI failed, maybe this needs to be updated?

boldtrn:
> I was wondering, why the CI checks did not run automatically. Can I somehow trigger the CI?

I think your first PR needs to get merged, before that your CI runs need to be manually approved. So next PR should work automatically 👍 .

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1115,45 +1115,45 @@ open class NavigationMapView: MGLMapView, UIGestureRecognizerDelegate {
 @objc(MBNavigationMapViewDelegate)
```
boldtrn, line 1115: This should be MLNavigationMapViewDelegate as well?

boldtrn: Probably the same should apply to MBNavigationMapViewCourseTrackingDelegate?

boldtrn (changes requested):
Very nice PR, thank you very much for working on this. Would you mind writing a Changelog entry, describing what you changed and what implementers need to change?

Also could you check the naming issue I commented on?

MapboxNavigation/NavigationMapView.swift
mindthefish: Ok, that slipped through. The same problem exists in countless other places. Where I'm not sure: We have swapped the prefix `MGL` for `MLN` SDK-wide. But for the @objc attribute the prefix `MB` was used. Should we then change this to `ML` or align the prefix and use `MLN` here as well?

boldtrn: I think it would make sense to swap out MBGL to MLN. We could do this in this PR or do this in a follow-up. I think only renaming a smaller part would be a bit weird.

I would be fine with merging this PR sooner rather than later and do a second renaming, maybe after #24 is merged as well to avoid too many conflicts?

mindthefish: `MGL` to `MLN` is not the problem. I was referring to the `@obj` attributes which have the prefix `MB` with 98 occurencies in 41 files:

image

Is there a problem renaming them to `MLN` as well? 

We can also wait until https://github.com/maplibre/maplibre-navigation-ios/pull/24 is merged, so I can resolve the conflicts before merging this one.

boldtrn (approved):
LGTM 👍  Thanks for working on this.

boldtrn:
Ok, I think let's go ahead with merging this, I will create a follow-up for renaming the `@objc` annotations.

PR #31, opened 2024-03-15
Configure first voice instruction.

StephanPartzsch:
With the new parameter `speakFirstInstructionOnFirstStep` in `RouteController` the first instruction on the first step can be omitted.


boldtrn (approved):
LGTM 👍

PR #32, opened 2024-03-16
Let speak instructions without route progress.

StephanPartzsch:


MapboxNavigation/RouteVoiceController.swift
```diff
@@ -209,29 +205,29 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
...
-        
-        if #available(iOS 10.0, *), utterance?.voice == nil {
+        if 
+            #available(iOS 10.0, *),
+            !ignoreProgress,
+            utterance.voice == nil, 
+            let routeProgress
+        {
+            let modifiedInstruction = voiceControllerDelegate?.voiceController?(self, willSpeak: instruction, routeProgress: routeProgress) ?? instruction
+
```
boldtrn, line 221: I think you introduce an error here.

The voiceControllerDelegate should be called even if utterance.voice != nil or if iOS 10.0 isn't available. This was done before.

This function is a bit messy here. Essentially this should allow implementers to change details about the instruction before speaking it.

boldtrn: I think it is correct that just for speaking an instruction, we might not necessarily require a routeProgress. In this case it might be fine to simply not create the modified instruction or set the instruction that was initially passed in this function as modifiedInstruction and then follow the regular flow.

boldtrn (changes requested):
Please see my comment, I think your proposed version is not correct yet.

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -264,8 +257,8 @@ public protocol VoiceControllerDelegate {
      
      - parameter voiceController: The voice controller that will speak an instruction.
      - parameter instruction: The spoken instruction that will be said.
-     - parameter routeProgress: The `RouteProgress` just before when the instruction is scheduled to be spoken.
+     - parameter routeProgress: The `RouteProgress` just before when the instruction is scheduled to be spoken. Could be `nil` if no progress is available or if oit should be ignored.
```
boldtrn, line 260:
```suggestion
     - parameter routeProgress: The `RouteProgress` just before when the instruction is scheduled to be spoken. Could be `nil` if no progress is available or if it should be ignored.
```

Minor typo

Examples/Swift/ViewController.swift
```diff
@@ -360,7 +360,7 @@ extension ViewController: VoiceControllerDelegate {
         print(interruptedInstruction.text, interruptingInstruction.text)
     }
 
-    func voiceController(_ voiceController: RouteVoiceController, willSpeak instruction: SpokenInstruction, routeProgress: RouteProgress) -> SpokenInstruction? {
+    func voiceController(_ voiceController: RouteVoiceController, willSpeak instruction: SpokenInstruction, routeProgress: RouteProgress?) -> SpokenInstruction? {
```
boldtrn, line 363: Could you please add a CHANGELOG entry for the changed APIs? Both here and for the RouteVoiceController?

MapboxNavigation/RouteVoiceController.swift
```diff
@@ -192,13 +192,9 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
...
      - parameter locale: The `Locale` used to create the voice read aloud the given instruction. If `nil` the `Locale.preferredLocalLanguageCountryCode` is used for creating the voice.
+     - parameter ignoreProgress: A `Bool` that indicates if the routeProgress is added to the instruction.
      */
-    open func speak(_ instruction: SpokenInstruction, with locale: Locale?) {
-        guard let routeProgress else {
-            assertionFailure("routeProgress should not be nil.")
-            return
-        }
-        
+    open func speak(_ instruction: SpokenInstruction, with locale: Locale?, ignoreProgress: Bool = false) {
```
boldtrn, line 197: To be honest, I am not sure if we really need the ignoreProgress switch here? 

We would need to make route progress optional as well for `modifiedInstruction.attributedText(for: routeProgress.currentLegProgress)`.

boldtrn (changes requested):
We really need a CHANGELOG entry if we change public APIs

boldtrn (approved):
Thanks, LGTM 👍

PR #33, opened 2024-04-04
Fix CI for Swift Package Manager

Patrick-Kladek:
Creating a PR here in hopes the CI will run.

This PR was done to finalize the work from #24. As the original Author is not responding we continue the official discussion here.

PR #37, opened 2024-04-12
Consistent Code Style

Patrick-Kladek:
### Description

As discussed in #33 we want to ensure a consistent code style. Therefore this PR adds SwiftLint and also SwiftFormat.

### Tasks

- [x] Add Swiftlint Plugin (this is a precompiled binary to reduce build time)
- [x] Add SwiftFormat, and let it auto configure and save results
- [x] Run SwiftFormat automatically on commit + add Documentation how to enable it.

### Infos for Reviewer

SwiftLint requires iOS13 and above, if that's a deal breaker I'll remove SwiftLint

Had to disable the following default rules as they were producing a lot of errors which need some re-architecture and while doing so might break something:
  - force_cast
  - function_body_length
  - identifier_name
  - cyclomatic_complexity
  - type_body_length
  - file_length
  - force_try

Now that SwiftLint runs again, how should we deal with the TODO annotations in the codebase that SwiftLint exposes?


Patrick-Kladek:
@StephanPartzsch @boldtrn As discussed earlier this PR reformats the whole codebase. Can you please review it

.github/workflows/ios.yml
```diff
@@ -19,10 +19,10 @@ jobs:
...
+        uses: actions/checkout@v4
       - name: Build
         run: |
-          set -o pipefail && xcodebuild build -scheme maplibre-navigation-ios -destination 'platform=iOS Simulator,OS=17.0.1,name=iPhone 15 Pro' | xcbeautify --renderer github-actions
+          set -o pipefail && xcodebuild build -scheme maplibre-navigation-ios -destination 'platform=iOS Simulator,OS=17.0.1,name=iPhone 15 Pro' -skipPackagePluginValidation | xcbeautify --renderer github-actions
       - name: Test
         run: |
-          set -o pipefail && xcodebuild test -scheme maplibre-navigation-ios -destination 'platform=iOS Simulator,OS=17.0.1,name=iPhone 15 Pro' | xcbeautify --renderer github-actions
\ No newline at end of file
+          set -o pipefail && xcodebuild test -scheme maplibre-navigation-ios -destination 'platform=iOS Simulator,OS=17.0.1,name=iPhone 15 Pro' -skipPackagePluginValidation | xcbeautify --renderer github-actions
```
boldtrn, line 28: It's a bit unfortunate that we need this now, but OK, I guess this is required for swiftlint

.swiftformat
boldtrn, line 1: Out of interest, where does this file come from? Is this like the default config?

Patrick-Kladek: you can run this command to analyze the project and it will store the configuration.

```bash
swiftformat --inferoptions . --output .swiftformat
```

https://github.com/nicklockwood/SwiftFormat?tab=readme-ov-file#config-file

.tools/git-format-staged-LICENSE.txt
boldtrn, line 1: great that you remembered to add the license as well 👍 - I guess adding MIT licensed source code should be fine, just checking with @wipfli if this is fine for MapLibre.

CONTRIBUTING.md
```diff
@@ -15,13 +15,21 @@ When reporting a bug in the navigation SDK itself, please indicate:
 
 ## Building the SDK
 
-To build this SDK, you need Xcode 14 and [Carthage](https://github.com/Carthage/Carthage/):
+To build this SDK, you need Xcode 15 and [Carthage](https://github.com/Carthage/Carthage/):
```
boldtrn, line 18: Carthage is not required anymore :)

Package.swift
```diff
@@ -3,103 +3,108 @@
...
-				//
-				//	~/Library/Developer/Xcode/DerivedData/maplibre-navigation-ios-cdijqyqwjamndzfaqhxchbiayzsb/Build/Products/Debug-iphonesimulator/maplibre-navigation-ios_MapboxNavigationTests.bundle: bundle format unrecognized, invalid, or unsuitable
-				//	Command CodeSign failed with a nonzero exit code
-				//
-				// Instead the json files are placed in a Folder called 'Fixtures' and manually specified for copying
-				// The Assets.xcassets is compiled into an Assets.car
-				// This results in a flat Bundle file structure however the tests pass.
+    name: "maplibre-navigation-ios",
+    defaultLocalization: "en",
+    platforms: [.iOS(.v13), .macOS(.v10_15)],
```
boldtrn, line 8: Why the platform change?

.tools/git-format-staged-LICENSE.txt
boldtrn: Also @ianthetechie here, I'd just like a second opinion on this one 👍

Package.swift
Patrick-Kladek: My bad, I was experimenting with different ways of adding SwiftFormat, shouldn't be needed anymore

CONTRIBUTING.md
Patrick-Kladek: As we want to bring back the Example I thought to leave this in. If you think its better to remove it, I'll do

Patrick-Kladek:
SwiftLint Package requires iOS13, maybe we can raise an Issue there https://github.com/lukepistrol/SwiftLintPlugin?

Bildschirmfoto 2024-04-12 um 19 14 06

.github/workflows/ios.yml
Patrick-Kladek: I've used SwiftLintPlugin which only contains the binary. We can of course always use SwiftLint repo and build it from source, I think this would eliminate this option, however it would significantly increase build time on our CI and also add various unrelated packages (like SourceKitten, ...) to users of this package

StephanPartzsch:
Speaking of packages and SPM. Do we have in mind to add a line to the readme that SPM is now an additional option and not only carthage?

boldtrn:
@StephanPartzsch, Carthage has been removed actually.

CONTRIBUTING.md
boldtrn: Ah ok, so for the Examples we would need Carthage?

boldtrn:
> SwiftLint Package requires iOS13, maybe we can raise an Issue there https://github.com/lukepistrol/SwiftLintPlugin?

I general I think going to iOS13, if we have to is no problem. Doing this for tooling, I am not sure? Do we get enough benefit from SwiftLint that we believe this is worth it? Or should just run the auto format for now?

So SwiftLint introduces two possible issues so far:
- The build command with `-skipPackagePluginValidation`
- Raise iOS to 13

Patrick-Kladek:
@boldtrn I've submitted a PR which drops the min version for SwiftLint, till this is merged we can use my fork to keep iOS12 supported. 

I guess in September when Xcode 16 becomes mandatory, iOS13 becomes the last supported version, till then we can go ahead with my fork. Even if my changes are not accepted for half a year I can maintain the fork

.tools/git-format-staged-LICENSE.txt
ianthetechie: Yes, I believe that MIT licensed code is acceptable. Oliver may know better than me on whether MapLibre has a "license policy", but I would assume that any of the permissive licenses including (modified) BSD, MIT, ISC, and Apache 2.0 are acceptable. As far as I know, all of the above licenses are also fully compatible with each other.

.swiftformat
```diff
@@ -54,7 +54,7 @@
 --patternlet hoist
 --ranges spaced
 --redundanttype infer-locals-only
---self remove
+--self insert
```
Patrick-Kladek, line 57: Maybe a controversial change but I think it makes sense.

Swift allows you to omit `self` when using local properties or functions. This is great in Xcode where you have syntax highlighting so you can easily differentiate between local properties/functions or global ones.

This context is completely lost in GitHub where code review happens. To make it easier for reviewers I would propose to enforce `self` usage.

Whats your take on this, is this fine or should I revert this?

.tools/git-format-staged-LICENSE.txt
Patrick-Kladek: @wipfli can you help with this question please?

CONTRIBUTING.md
Patrick-Kladek: Good point, we shouldn't need Carthage at all, therefore I've removed this. I think for the Examples it should be enough to add a new Xcode project and add the local Package as dependency.

.tools/git-format-staged-LICENSE.txt
```diff
@@ -0,0 +1,7 @@
+Copyright 2018 Jesse Hallett
```
wipfli, line 1:
```suggestion
Copyright 2024 MapLibre contributors
Copyright 2018 Jesse Hallett
```

.tools/git-format-staged-LICENSE.txt
wipfli: If I understand correctly, the question is if it is OK to add `git-format-staged`, an MIT-licensed python script, to the source tree if we include the license file?

The answer to that is, yes, you can include the file as long as you keep the license.

I guess you could add a line at the top of the license  mentioning MapLibre contributors. But if you don't change the code, I think it is also not necessary to add that line...

.swiftformat
boldtrn: I don't have strong feelings on this one. The default seems to be to remove self. If in doubt, I'd say, let's stick with the defaults unless we have a good reason not to follow defaults. So I am leaning towards `--self remove` or not to mention the parameter at all and fall back to the default?

.github/workflows/ios.yml
boldtrn: Sorry for being a bit hesitant here. Could we maybe step one step back and discuss if we need the SwiftLintPlugin. I am still not 100% sure we need this plugin.

For CI we could install SwiftLint through brew or we could install it through SPM. Would this change that much about the CI build time?

While looking at other repos that use SwiftLint I stumbled over the repo from @ianthetechie here: https://github.com/stadiamaps/maplibre-swiftui-dsl-playground

They seem to use the brew install method. I am open to discussing different options, but if we can use widely used standard tools in favour of a fork of a plugin, that requires also to change our build settings, then I am leaning towards going down the standard route.

.swiftformat
Patrick-Kladek: My reasoning here was to improve readability when reviewing in GitHub. Is this not a strong enough reasoning for you?

.tools/git-format-staged-LICENSE.txt
Patrick-Kladek: We are not changing the code of the python script `git-format-staged` only using it

.github/workflows/ios.yml
ianthetechie: Actually, funny that you notice that @boldtrn. A few comments...

1. We actually use swiftformat rather than swiftlint. These are quite similar projects. I don't actually have much of an opinion on preferring one or the other (or both) for a project. We just happen to use swiftformat.
2. Aside: swiftformat is actually installed in the base GitHub runners. The main reason we run `brew install` is because we have no idea when GitHub will update it 😂 So to simplify things, we just have an internal policy of running the latest and using `brew install` in Actions.
3. I agree I'd prefer to minimize use of plugins and (this is *purely* personal preference) tend to go for CI checks over commit hooks as they are significantly more robust. Mostly this is because it's *usually* simpler, more robust, and has fewer potential security risks. Just my opinion.

Patrick-Kladek: I have worked in multiple companies now and it was de-facto standard to use SwiftLint. The reason I've added SwiftLint to Xcode and not actions was, that you see the warnings right away, instead of waiting for the CI to finish.

SwiftFormat can work in tandem with SwiftLint which is even encouraged, as it can automatically fix issues, the more complex ones needs to be fixed by the programmer.

If we move this to CI, how are we communicating back the warnings it found? Thats an unsolved issue right now, you would need to manually go to the checks tab and then click on summary to see the warnings - they don't show up in the PR. 

Bildschirmfoto 2024-04-17 um 18 09 12

Of course we can fail the build if a warning is detected, but for this we first need to resolve the existing ones, where we also don't have a clear path. I've already silenced quite a lot of warnings SwiftLint found as they need quite big changes which we want to do in separate PRs anyway.

Patrick-Kladek:
SwiftLint Plugin was removed

PR #41, opened 2024-05-13
Use Map without Navigation

Patrick-Kladek:
### Description

We need to start a navigation from an existing map, without presenting another map as overlay. To achieve this we need to show a normal map first and then assign a route which will start navigation

### Open Tasks

- [x] make route parameter optional

### DIFF
--- a/MapboxNavigation/NavigationView.swift
+++ b/MapboxNavigation/NavigationView.swift
@@ -165,7 +165,7 @@ open class NavigationView: UIView {
     
     func commonInit() {
         self.setupViews()
-        setupConstraints()
+        self.setupConstraints()
     }
     
     func setupStackViews() {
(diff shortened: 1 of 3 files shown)


boldtrn:
Thanks for opening this PR. In general I like the idea of a "Navigation" without a route (we have this in our app as well).  I haven't tested your PR, but while scanning the code, it seems like the camera and puck won't work without the route?

Let's maybe discuss the expectation for this mode first:
So I would expect that the navigation without route would still focus around the user location and follow the user location while moving. Is this what you are aiming for?

Patrick-Kladek:
@boldtrn this is early work in progress, I made this PR to share my progress with my colleagues so its nowhere near working. In fact right now you can only initialize `NavigationViewController` without a Route, however none of the UI reflects that, camera is pointing to (0, 0) coordinates instead of current location, location puck is not appearing and so on.

> So I would expect that the navigation without route would still focus around the user location and follow the user location while moving. Is this what you are aiming for?

Yes thats exactly what I want. You start with a normal map, where you can tap POIs, see your current location and so on. Once you assign a route, it will switch to navigation mode where you will see the navigation UI, spoken instructions, ...

boldtrn:
Yeah no worries, I just saw it and figured I give a bit of feedback as well :)

PR #44, opened 2024-05-21
Fix failing test for 24 hr locales

michaelkirk:
### Description

This test was failing for me. I use `en-US` locale, but I set my clock to 24 hour. It's admittedly uncommon, but I'd wager this test is failing for anyone in a non-12 hours locale.

```
 let beforeSunrise = dateFormatter.date(from: "05:00 AM")! // Thread 1: Fatal error: Unexpectedly found nil while unwrapping an Optional value
```

To be honest, I expected the existing date format string `hh:mm a` to be sufficient, but apparently it's not.


boldtrn (approved):
Thanks for catching this 👍

PR #45, opened 2024-05-22
Deprecate some hardcoded mapbox styles, and replace those that remain with maplibre styles. (+docs!)

### LINKED ISSUE #42: NavigationViewController map view is blank (SwiftUI)
### Steps to Trigger Behavior

1. Set up example app:
``` 
git clone https://github.com/michaelkirk/maplibre-navigation-ios-example-app
cd maplibre-navigation-ios-example-app
git checkout mkirk/navigation-view-controller-blank
```
2. Update ContentView with your accessToken
3. run app

https://github.com/michaelkirk/maplibre-navigation-ios-example-app/tree/mkirk/navigation-view-controller-blank

### Expected Behavior

I expect some kind of map to be rendered.

### Actual Behavior

Other UI is rendered, like the controls and user location puck, but the base map is just solid gray.

Screenshot 2024-05-14 at 19 34 04

### Version(s) affected

- Platform: iOS
- OS version: iOS 17.4 
- Device type: iPhone 5 Simulator and iPhone SE (2020), 17.4, physical device
- Emulator/ Simulator: yes
- Development OS: macos 14.4.1
- maplibre-navigation-ios Version: maplibre-navigation-ios main (3eda169)
- MapLibre GL version [e.g. 6.3.0]: MapLibre Native 6.4.1

### Additional context

I'm using the NavigationViewController in SwiftUI [via `ViewControllerRepresentable`](https://github.com/michaelkirk/maplibre-navigation-ios-example-app/blob/mkirk/navigation-view-controller-blank/MapLibreNavigationExampleApp/ContentView.swift#L73). I haven't tried it in a non Swift UI app yet.

I can render a `MapView` and `NavigationMapView` just fine on it's own (via `ViewRepresentable`). It seems to only be an issue in the context of the `NavigationViewController`:

Screenshot 2024-05-14 at 19 52 33

michaelkirk:
Fixes #42 

### Description

Following up on https://github.com/maplibre/maplibre-navigation-ios/issues/42#issuecomment-2114511460

In the mapbox days, it arguably made sense to have hardcoded defaults that referenced mapbox servers. I think we can agree that in the maplibre world it no longer makes sense to reference these mapbox services.

Perhaps more controversially: Post maplibre, I think it makes less sense to have any default service provider at all. Instead, we should surface API's that require the implementor to pass in an explicit backend tile provider. I've introduced some new API's to that end, and deprecated some of the existing API's which utilize a default tile source.

## Screenshot of previous API along with the newly proposed deprecation messages
Screenshot 2024-05-22 at 17 16 56

For example: 
```swift
let styleURL = AppConfig().tileserverStyleUrl

// previous broken way
// `styles` were optional. The default value specified mapbox, which was broken.
let vc = MapboxNavigation.NavigationViewController(
    for: route,
)

// previous way that works, but is now proposed to being deprecated, because `styles` isn't actually optional in practice.
let style = DayStyle()
style.mapStyleURL = styleURL
let vc = MapboxNavigation.NavigationViewController(
    for: route,
    styles: [style]
)

// new way 1: pass in non-optional style
let style = DayStyle(styleURL: styleURL)
let vc = MapboxNavigation.NavigationViewController(
    for: route,
    dayStyle: dayStyle
)

// new way 2: pass in only URL, using default colors for night and day UI components
let vc = MapboxNavigation.NavigationViewController(
    for: route,
    dayStyleURL: styleURL
)
```

We still do make available some "demo" styles that use https://demotiles.maplibre.org, but I've intentionally made it explicit that it's a demo - `DayStyle(demoStyle: ())`. See below for the alternatives rationale.

### Swap out mapbox tile server for malibre tile server

I could have maintained the current APIs where specifying styles is optional, and simply swapped the mapbox tiles out for maplibre demo tiles. However, my understanding of the maplibre project is, unlike mapbox, maplibre doesn't (currently anyway) intend to host production map services, and so maplibre libraries should direct people towards finding their own service providers, even at some cost to the "everything just works" experience. I think providing an easy way to use the demo services for POC, while trying to nudge users to proper external tile hosts is the right balance. But I'm open to learning more about this if people feel differently.

Another alternative that would be a  smaller code change would be to keep the existing init, but make `styles` in `NavigationViewController.init(styles:)` non-optional. This would still be a semver breaking change, but an arguably simpler change for users to adapt to. My understanding is that, in practice, the user would pass in 0, 1, or 2 styles. Accepting an array of 3 or more styles is not useful as far as I know. Plus I think the old API where "the second style should be the night style" is a bit obscure. Since we were breaking the API anyway, I decided to codify the expectations.

### Info for reviewers

I'm very new to the maplibre native ecosystem, so consider this PR a proposal open for comment and changes. I look forward to your feedback.


MapboxNavigation/CarPlayMapViewController.swift
```diff
@@ -50,8 +50,8 @@ class CarPlayMapViewController: UIViewController, MLNMapViewDelegate {
         super.viewDidLoad()
 
         self.styleManager = StyleManager(self)
-        self.styleManager.styles = [DayStyle(), NightStyle()]
-        
+        self.styleManager.styles = [DayStyle(demoStyle: ()), NightStyle(demoStyle: ())]
```
michaelkirk, line 53: The demo styles are now being used in some places that were previously defaulting to mapbox styles. This is probably wrong, but (I think) no worse than things being completely broken before.

Probably these should somehow be surfaced or inferred as well. I somewhat arbitrarily decided this was out of scope for this PR, but I could follow up in this PR or another depending on how people are feeling about it.

michaelkirk:
Whoops, converting to a draft while I address a missed TODO.

MapboxNavigationTests/Sources/Tests/NavigationViewControllerTests.swift
```diff
@@ -87,13 +89,18 @@ class NavigationViewControllerTests: XCTestCase {
         XCTAssertFalse(wayNameView.isHidden, "WayNameView should be visible.")
     }
     
+    // TODO: This test doesn't seem to be testing anything. Adding a test where `updatedStyleNumberOfTimes = 1` might be illuminating
     func testNavigationShouldNotCallStyleManagerDidRefreshAppearanceMoreThanOnceWithOneStyle() {
```
michaelkirk, line 93: These tests seem a little... weird. I left some comments expressing my confusion, but opted not to try to change them more than necessary in this PR. I don't think I made them any more weird.

MapboxNavigation/Style.swift
```diff
@@ -34,28 +34,52 @@ open class Style: NSObject {
...
+    }
+
+    @objc public convenience init(demoStyle: ()) {
+        self.init(mapStyleURL: MLNStyle.defaultStyle().url)
+    }
+}
+
+extension Style: NSCopying {
+    public func copy(with zone: NSZone? = nil) -> Any {
+        let copy = Self(mapStyleURL: self.mapStyleURL)
```
michaelkirk, line 74: Living in Swift land for a while, I forgot about NSCopying!

michaelkirk:
Re-marking this as ready for review. 

I brushed up against some things that could use more attention (CarPlay and weird NavigationViewController tests). I didn't fix everything, but I think everything changed is either neutral or an improvement.

MapboxNavigationTests/Sources/Tests/NavigationViewControllerTests.swift
boldtrn: Thanks, yeah, some of the tests are indeed weird. We should be fine to improve them later on 👍

boldtrn (changes requested):
Thanks for working on this. Could you add a Changelog entry explaining what you changed?

MapboxNavigationTests/Sources/Tests/NavigationViewControllerTests.swift
boldtrn: I am not sure if we should add all the TODOs in the code though? Maybe you could summarise your findings in an issue and we can look at what happens exactly?

MapboxNavigation/CarPlayMapViewController.swift
boldtrn: I personally don't mind your approach here 👍

MapboxNavigation/DayStyle.swift
```diff
@@ -45,13 +45,12 @@ private extension UIFont {
...
 open class DayStyle: Style {
-    public required init() {
-        super.init()
-        mapStyleURL = MLNStyle.navigationGuidanceDayStyleURL
+    @objc public required init(mapStyleURL: URL) {
+        super.init(mapStyleURL: mapStyleURL)
         styleType = .day
         statusBarStyle = .default
     }
-
```
boldtrn, line 54: I am wondering about this change. Don't we have auto formatting enabled that should make sure we don't have these kinds of changes? Did you enable the auto formatting? See: https://github.com/maplibre/maplibre-navigation-ios/blob/main/CONTRIBUTING.md#enable-automatic-code-formatting

boldtrn: BTW: I don't really mind this particular change, I just expected that these types of things don't happen anymore.

CHANGELOG.md
```diff
@@ -12,6 +12,13 @@
  - Updated MapLibre Native dependency to ios-v6.0.0 (https://github.com/maplibre/maplibre-native/releases/tag/ios-v6.0.0). Implementers need to change the prefix MGL to MLN for all MapLibre Native classes that are referenced.
  - Only snap location to route if the location is within the `RouteControllerUserLocationSnappingDistance`
  - Add support for Swift Package Manager while dropping Carthage and Cocoapods.
+ - Removed implicit default dependencies on MapBox tileservers by requiring explicit styles URLs in more places.
+   - Merged in <https://github.com/maplibre/maplibre-navigation-ios/pull/45>.
+   - BREAKING: Removed `MLNStyle` extensions referencing non-functioning MapBox styles, e.g. `MLNStyle.navigationGuidanceDayStyleURL`.
```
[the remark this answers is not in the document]
boldtrn: Breaking is fine for me 👍

boldtrn (approved):
Thanks a lot for this change 👍

Issue #46, opened 2024-05-24
Strange things in NavigationViewControllerTests that seem not to be testing anything.


michaelkirk:
There are a few tests that seem to be in pursuit of making sure the stylemanager switches style when appropriate (and only when appropriate).

I noted this while working on some adjacent code, and [it was requested](https://github.com/maplibre/maplibre-navigation-ios/pull/45#discussion_r1610803604) I make an issue with my findings.

Weird things:

1. There's a test about tunnel travel, but it doesn't look like the route goes through any tunnel. 

2.  strange movement simulation.
I'm assuming the following code was intended to simulate travel by updating the location along the route:
```
routeController.locationManager(routeController.locationManager, didUpdateLocations: [someLocation])
routeController.locationManager(routeController.locationManager, didUpdateLocations: [someLocation])
routeController.locationManager(routeController.locationManager, didUpdateLocations: [someLocation])
```

But why is it using the same location all three times? I'd expect each call to be a further point along the route.

3. There are several tests that check that the style has *not* changed after certain actions, but the actions seems incoherent (see point 2.), and there is no test to checks the positive condition that the  style has ever changed, so I'm not confident that any of the expected code is being tested.

4. Some variables have misleading names - they seem to have nothing to do with actual intersections along the route. There's no taylorStreet, turkStreet, or fultonStreet at these points. I wonder if the underlying route changed at some point?  It's a minor thing but it added to my confusion.
```swift
let taylorStreetIntersection = routeController.routeProgress.route.legs.first!.steps.first!.intersections!.first!
let turkStreetIntersection = routeController.routeProgress.route.legs.first!.steps[3].intersections!.first!
let fultonStreetIntersection = routeController.routeProgress.route.legs.first!.steps[5].intersections!.first!
```

PR #47, opened 2024-05-24
Fix NavigationViewController rerouting

michaelkirk:
(**edit:** Sorry for the initial description. I had a CLI misfire.)

~~This is a draft because: ~~
- [x] This is based on #45, so please merge that first.
- [x] Looking for advice on testing the generated objc headers. - From what I can tell, objc build is not currently supported anyway.

The `reason` parameter was added in https://github.com/maplibre/maplibre-navigation-ios/commit/c65c9c64c7acd371e73fb7914f95657832aa42fa#diff-9644714366824e368806302bce45047e92c0c6db14650897cdb254146f8df15cL56, but it was not added to existing callers.

Since it's an `optional` delegate method, no compiler error was triggered, the old method just sat around not being run.

### DIFF
--- a/MapboxCoreNavigation/SimulatedLocationManager.swift
+++ b/MapboxCoreNavigation/SimulatedLocationManager.swift
@@ -35,16 +35,9 @@ private class SimulatedLocation: CLLocation {
     /// ```
     /// extension MyDelegate: SimulatedLocationManagerDelegate {
     ///     func simulatedLocationManager(_ simulatedLocationManager: SimulatedLocationManager, locationFor originalLocation: CLLocation) -> CLLocation {
-    ///         // go off track by 10 meters east
-    ///         let offsetByMeters = 10
-    ///         let bearing = 90
-    ///
-    ///         let distRadians = meters / 6_372_797.6 // earth radius in meters
-    ///         let lat1 = location.coordinate.latitude * .pi / 180
-    ///         let lon1 = location.coordinate.longitude * .pi / 180
-    ///         let lat2 = asin(sin(lat1) * cos(distRadians) + cos(lat1) * sin(distRadians) * cos(bearing))
-    ///         let lon2 = lon1 + atan2(sin(bearing) * sin(distRadians) * cos(lat1), cos(distRadians) - sin(lat1) * sin(lat2))
-    ///         return CLLocation(latitude: lat2 * 180 / .pi, longitude: lon2 * 180 / .pi)
+    ///         // go off track 100 meters East of the original route
+    ///         let offsetCoordinate = originalLocation.coordinate.coordinate(at: 100, facing: 90)
+    ///         return CLLocation(latitude: offsetCoordinate.latitude, longitude: offsetCoordinate.longitude)
     ///     }
     /// }
     /// ```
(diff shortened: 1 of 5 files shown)


MapboxNavigation/NavigationViewController.swift
```diff
@@ -618,7 +618,7 @@ extension NavigationViewController: RouteControllerDelegate {
         self.delegate?.navigationViewController?(self, willRerouteFrom: location)
     }
     
-    @objc public func routeController(_ routeController: RouteController, didRerouteAlong route: Route) {
```
michaelkirk, line 621: I think nested types like this aren't actually accessible to objc. 

I can move the type, but I wanted to be able to test it. I was wondering if anyone had advice for quickly testing the objc build of the library, or if I should just bit the bullet and set up a purpose built objc project.

MapboxCoreNavigation/SimulatedLocationManager.swift
```diff
@@ -40,7 +40,10 @@ open class SimulatedLocationManager: NavigationLocationManager {
      Specify the multiplier to use when calculating speed based on the RouteLeg’s `expectedSegmentTravelTimes`.
      */
     @objc public var speedMultiplier: Double = 1
-    
+
+    /// Instead of following the given route, go slightly off route. Useful for testing rerouting.
+    @objc public var goOffRoute: Bool = false
```
michaelkirk, line 45: I used this while testing and thought it might be useful to others. If you disagree, I can revert this portion - no big deal.

boldtrn: I think this is a nice idea, I am not sure about the implementation though. Would it be possible to have a delegate or something like that so implementers can decide how the offroute is calculated/offset?

I am open to how this is implemented, but some form of overwriting a method might be a good solution IMHO.

MapboxNavigation/NavigationViewController.swift
boldtrn: To be honest, I have no experience with objc. So I can't really help here.

boldtrn: Given that this existed previously, I would be fine with just leaving it as it is right now until someone has an actual issue and we can always fix this later?

boldtrn:
Thanks for catching this. It would be great if you could rebase your PR 👍

MapboxNavigation/NavigationViewController.swift
michaelkirk: I've done some objc and swift development, but I'm new to SPM. I haven't tried too hard, but I wasn't immediately able to figure out how to use this library from objc after installing from spm.


It's fine by me to leave it as is for now

boldtrn (approved):
LGTM 👍

boldtrn:
BTW: do you think, it would be possible to add a test case that machtes the updated reroute method? We have a assertFalse testcase, but maybe we could also add a asserTrue test case?

michaelkirk:
> BTW: do you think, it would be possible to add a test case that machtes the updated reroute method? We have a assertFalse testcase, but maybe we could also add a asserTrue test case?

Good idea, considering I was complaining about the same thing in https://github.com/maplibre/maplibre-navigation-ios/issues/46 😄. I've added a positive test case now.

In the process, I learned about Turf's `CLCordinate2D.coordinate(at:facing:)`, so I used it to simplify the `SimulatedLocationManagerDelegate` example code a bit.

PR #49, opened 2024-05-31
Fix simulated speed label in NavigationViewController

michaelkirk:
### Description

For testing purposes, there is a `SimulatedLocationManager`. You give it a route and it updates your "current location" to be along that route. 

To speed up the testing cycle, you can set a speed multiplier on the `SimulatedLocationManager` so that the route is travelled much faster than what it would be in realtime.

When using the `SimulatedLocationManager` the navigation view controller shows what the speed multiplier is. However, it used to always say "Simulating Navigation at 1x" regardless of what you'd set the speed multiplier to.

**before**
Screenshot 2024-05-31 at 10 44 26

Now it properly reflects the specified speed
**after**
Screenshot 2024-05-31 at 10 44 02

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,9 @@
    - Deprecated `DayStyle()`/`NightStyle()` initializers because they were backed by an implicit tile service. If these default styles *are* still used, they'll now use the MapLibre demo style.
    - `NavigationViewController` now expects explicit style URLs with `NavigationViewController(route:dayStyleURL:nightStyleURL:...)` or NavigationViewController(route:dayStyle:nightStyle:...)` and the existing initializer, which allowed "default" styles, is deprecated and uses the MapLibre demo styles.
  - Fix: NavigationViewController was not re-routing when the user went off route.
-   - Merged in <https://github.com/maplibre/maplibre-navigation-ios/pull/47>.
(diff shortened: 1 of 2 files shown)


michaelkirk:
I think I misread the test failure. The benchmark was marked as slower, but it didn't cause the tests to fail.

The failure is a cryptic:

```
Notice: Tests Passed: 0 failed, 0 skipped, 43 total (8.958 seconds)
Error: Process completed with exit code 65.
```

So tests passed, but non-zero exit code from the test runner.

The runs that fail seem to run much slower (e.g. more than 20+ minutes vs. the passing tests 5-10min)

Re-starting the action succeeded. I'm not sure what to make of it.

hactar (approved):
I approve this change - but I'm not going to merge it - someone with more knowledge about the tests should take a look and confirm that we're ok ☺️

boldtrn:
> Also @boldtrn usually would like people to add info about the PR to the changelog file ☺️

Just to clarify, while it certainly doesn't hurt, IMHO we don't need to document tiny changes like this. But especially if we remove code, change APIs, or change anything else, that might break / interrupt other implementers, we should really add this to the changelog.

Thanks for working on this @michaelkirk, I think we can merge this :)

PR #51, opened 2024-05-31
Fix broken keyboard layout after changing styles

### LINKED ISSUE #50: Setting styles can break keyboard/sheet layout (via `forceRefreshAppearance`)
Firstly, I'm not sure if this is a bug in maplibre, a bug in SwiftUI, or something in the interaction between.

I'm experiencing a problem in my own app which I've narrowed down to this method in maplibre-navigation-ios:

```
func forceRefreshAppearance() {
    for window in UIApplication.shared.windows {
        for view in window.subviews {
            view.removeFromSuperview()
            window.addSubview(view)
        }
    }
    
    self.delegate?.styleManagerDidRefreshAppearance?(self)
}
```

It's called when the Style is updated. `Style` determines map rendering but also UIKit widgets via UIAppearance.

I'm assuming the motivation for the above implementation is from https://developer.apple.com/documentation/uikit/uiappearance :

> iOS applies appearance changes when a view enters a window, but it doesn’t change the appearance of a view that’s already in a window. To change the appearance of a view that’s currently in a window, remove the view from the view hierarchy and then put it back.

Note that we're removing views from not just our own application windows, but *all* windows (including the keyboard window).  I've fixed this problem in my app by rework this method to ignore the keyboard window. (PR forthcoming)

### Steps to Trigger Behavior

I think it's something like:

1. add a map and an input field to the view hierarchy
2. make input field first responder to present keyboard
3. present another MapView modally
4. dismiss the modal mapview
5. make input field first responder again

You can see the steps to reproduce in my app here:

https://github.com/maplibre/maplibre-navigation-ios/assets/217057/4a75e8e3-475c-48d5-87ba-c876f17eacd7

I'm sorry I don't have a good minimal reproduction yet

### Expected Behavior

Presenting another mapview should not cause layout issues with keyboard presentation.

### Actual Behavior

When the keyboard is presented, the view layout gets real wonky, making most of my views non-visible.

Screenshot 2024-05-31 at 15 44 42

Should look like this:

Screenshot 2024-05-31 at 15 49 32

I suspect it's related to keyboard presentation/offset. If I hide/show the keyboard (cmd+k) you can see the view go from normal looking to broken:

https://github.com/maplibre/maplibre-navigation-ios/assets/217057/c3dfb563-f578-4e74-81bc-721bf596ae43

Note these videos are from the simulator, but I've reproduced the issue on a physical device as well.

### Version(s) affected

- Platform: iOS
- OS version: tested on 16.4 (simulator), 17.5.1 (physical device: iPhone SE)
- Device type: [e.g. iPhone6] iPhone SE
- Emulator/ Simulator: Reproduced on both simulator and device
- Development OS: 14.5 (23F79)
- maplibre-navigation-ios Version [e.g. 7.0.9]: 3ffaa51d (3.0.0-beta.1)
- MapLibre GL version [e.g. 6.3.0]:  6.4.1

### Additional context

I suspect that this problem might only manifest if you change styles when a keyboard is already presented. Normally this might not happen if you're using just one map, but in my case I have a first map for searching and finding a destination, and then modally present a second map for navigation purposes (I'm following #41 in hopes of getting rid of this second MapView). 

The initialization of the map sets `styles` which is considered a "change", and fires the above problematic method that removes all views from all windows.

michaelkirk:
FIXES #50 

I'm assuming the motivation for the previous implementation is from
https://developer.apple.com/documentation/uikit/uiappearance :

> iOS applies appearance changes when a view enters a window, but it
> doesn’t change the appearance of a view that’s already in a window. To
> change the appearance of a view that’s currently in a window, remove the
> view from the view hierarchy and then put it back.

Note that we're removing views from not just our own application
windows, but all windows (including the keyboard window). By ignoring
the keyboard window, my layout issues go away.

Previous steps to reproduce:

1. add a map and an input field to the view hierarchy
2. make input field first responder to present keyboard (thus adds
   keyboard window to application)
3. present another MapView modally (triggers `applyStyle`, which in turn
   re-adds every view from every window.
4. dismiss the modal mapview
5. make input field first responder again

At this point, the views in my application all seem to be inexplicably
offset by the keyboard height.

## Note there are a couple things I haven't tested yet:

- [ ] I don't use carplay.
- [ ] I don't have a night theme.

I'll eventually add those things to my own app and be able to test them, but it might be some weeks. In the meanwhile, I thought it'd be good to have this PR available in case someone else hits this problem. It took me a while to track it down.


boldtrn:
Thanks for looking into this. Yeah I think especially dark mode could potentially be an issue here. I agree that we should do some more testing on this. What I don't understand from your video, why is the keyboard suddenly open after the navigation? Because the keyboard was closed before starting the navigation. Is this intended? If yes, this feels quite exotic. (I am not saying, we shouldn't merge this, I'd just like to understand this a bit better)

michaelkirk:
> What I don't understand from your video, why is the keyboard suddenly open after the navigation

The keyboard is not suddenly open after navigation. It remains dismissed (12s-14s) until the user taps into the field (15s).

https://github.com/maplibre/maplibre-navigation-ios/assets/217057/4a75e8e3-475c-48d5-87ba-c876f17eacd7

I don't know the private implementation details of how iOS does keyboard presentation, but here's some debugging info that might be relevant:

Originally the app has just one UIWindow:

image


When the keyboard is first presented a UITextEffectsWindow is added:

image

After the input resigns first responder, the keyboard is dismissed, but the UITextEffectsWindow remains (though one of it's views is removed):

image

So that convinced me that even though the keyboard is not visible when the second MapView is dismissed, some of the keyboards containing views and windows have persisted.  With the previous logic, those views were being removed and re-added to their window. It seems plausible for iOS to say "don't touch views in our private windows" unless you want things to break.

So my theory is that, by removing/re-adding the containing keyboard views to their window, we've broken something. But the breakage is not apparent until the keyboard is actually shown.

PR #53, opened 2024-06-03
Fix #27: navigation puck falls off route when panning

### LINKED ISSUE #27: Moving the map with active navigation leads to marker getting dragged off the route
### Steps to Trigger Behavior

1. Start the example app and navigate
2. Move the map
3. The location marker is moved with the map and does not stay at the user's location
4. The location flies back on the route

### Expected Behavior

The marker should stay on the location of the user

### Actual Behavior

It is moved away

### Version(s) affected

- Platform: iOS
- OS version: iOS 17 (pretty sure it happens across different iOS versions)
- Emulator/ Simulator: both
- maplibre-navigation-ios Version current main
- MapLibre GL version 6.0.0

michaelkirk:
### Description

The bug was especially noticeable when panning quickly or zooming in
while navigating.

The issue is that the upstream method signature changed in
https://github.com/maplibre/maplibre-native/commit/76469a0c2227927f03a15fb95b3a20f34becc78c and then again in
https://github.com/maplibre/maplibre-native/commit/8ad5f9c1fd58c60b93d26c2819637fd8f2144fd1

So our "overridden method" wasn't actually being called.

It's not the commit authors fault that it broke - this is a private
method we're calling.

A longer term fix might be to include some kind of public stable
delegate method rather than an override.

## Before

https://github.com/maplibre/maplibre-navigation-ios/assets/217057/6240dec0-7e48-41b4-be0b-2e3b281257cf

## After

https://github.com/maplibre/maplibre-navigation-ios/assets/217057/9e6bcf5e-e25a-4fca-a241-1932bbf98f77

### Open Tasks

- Fixes #27

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -257,9 +257,9 @@ open class NavigationMapView: MLNMapView, UIGestureRecognizerDelegate {
             super.anchorPoint(forGesture: gesture)
         }
     }
-    
-    override open func mapViewDidFinishRenderingFrameFullyRendered(_ fullyRendered: Bool) {
-        super.mapViewDidFinishRenderingFrameFullyRendered(fullyRendered)
+   
+    override open func mapViewDidFinishRenderingFrameFullyRendered(_ fullyRendered: Bool, frameEncodingTime: Double, frameRenderingTime: Double) {
+        super.mapViewDidFinishRenderingFrameFullyRendered(fullyRendered, frameEncodingTime: frameEncodingTime, frameRenderingTime: frameRenderingTime)
         
         guard self.shouldPositionCourseViewFrameByFrame else { return }
         guard let location = userLocationForCourseTracking else { return }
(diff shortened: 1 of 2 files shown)


michaelkirk:
I just noticed there already *is* a delegate method: `mapViewDidFinishRenderingFrame` maybe we can use that instead? 

Converted to a draft while I look a bit more.

michaelkirk:
Since NavigationMapView is a subclass of MLNMapView, `NavigationMapView.delegate` is just `MLNMapView.delegate`. We probably can't utilize the delegate functionality within NavigationMapView because we want downstream users of `NavigationMapView` to be able to set their own mapview delegate.

It's often a bit tricky to mix inheritance and delegation. We want some kind of interceptor pattern.

I'm sure we could come up with something, but it seems like the solution will be less than trivial, so let's start by just updating the private header for now. (re-opening the PR as is)

boldtrn:
Thanks for looking into this @michaelkirk. @mindthefish as you saw serious issues with this, would you like to give this a try if this also fixes #27? I will also give this a try 👍

boldtrn (approved):
Thanks for looking into this, I can verify that this fixes the issue.

PR #54, opened 2024-06-06
Start & Stop Navigation in existing Map

Patrick-Kladek:
### Description

A long wanted feature was to interact with a map and later on start a navigation without presenting another map on top. This PR finally allows this. In retrospect it was quite simple, In `NavigationViewController` the route property is now optional and there are additional functions added to start & stop the navigation. 

### Open Tasks

- [x] Add Example Project to iterate faster
- [x] make route optional & add start/stopp methods
- [x] correctly show & hide UI for navigation
- [x] Remove Feedback form (from Mapbox, MapLibre doesn't have servers to send feedback)
- [x] Don't show EndOfRouteViewController after navigation finished, instead call delegate

### Infos for Reviewer

In the Example, the route calculation is done with Mapbox, users need a mapbox account and add the api-key in Secrets.xcconfig file. This file ignores changes, so the api-key should not leak into the git repo.

A nice side effect of this Task is, that we to got rid of some implicitly unwrapped optionals.

### Breaking Changes

* Feedback Button was removed, this button didn't work as it was a Mapbox feature and MapLibre doesn't have any servers. If this button is needed, it should be added manually.
* No End of Route View is presented on arrival. If it's needed, it should be presented manually in `navigationViewControllerDidFinish()` delegate method. The `EndOfRouteViewController` is still bundled in the project.


MapboxNavigation/NavigationViewController.swift
```diff
@@ -389,191 +379,137 @@ open class NavigationViewController: UIViewController {
...
-    public convenience init(for route: Route,
-                            dayStyleURL: URL,
+    @objc(initWithDayStyleURL:nightStyleURL:directions:routeController:locationManager:voiceController:)
+    public convenience init(dayStyleURL: URL,
                             nightStyleURL: URL? = nil,
                             directions: Directions = Directions.shared,
                             routeController: RouteController? = nil,
                             locationManager: NavigationLocationManager? = nil,
-                            voiceController: RouteVoiceController? = nil) {
+                            voiceController: RouteVoiceController = RouteVoiceController()) {
```
Patrick-Kladek, line 388: Right now we offer 3 or 4 different init options. This means you can still start a navigation the old way with presenting the sheet but this includes maintenance complexity.

I would prefer to only offer one - the new way. I'm aware that this will break compatibility, however this is a big feature that warrants a major version increase, so I think we should take the opportunity to cleanup. Whats your thoughts?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -389,191 +379,137 @@ open class NavigationViewController: UIViewController {
...
+        self.view.addSubview(mapSubview)
         mapSubview.pinInSuperview()
-        mapViewController.reportButton.isHidden = !self.showsReportFeedback
         
         self.styleManager = StyleManager(self)
         self.styleManager.styles = [dayStyle, nightStyle]
-
-        if !(route.routeOptions is NavigationRouteOptions) {
-            print("`Route` was created using `RouteOptions` and not `NavigationRouteOptions`. Although not required, this may lead to a suboptimal navigation experience. Without `NavigationRouteOptions`, it is not guaranteed you will get congestion along the route line, better ETAs and ETA label color dependent on congestion.")
-        }
```
Patrick-Kladek, line 471: most of the code moved to `startNavigation()` function below

MapboxNavigation/NavigationViewController.swift
```diff
@@ -628,11 +565,11 @@ extension NavigationViewController: RouteMapViewControllerDelegate {
         self.delegate?.navigationViewController?(self, didSelect: route)
     }
     
-    @objc public func navigationMapView(_ mapView: NavigationMapView, shapeFor routes: [Route]) -> MLNShape? {
+    public func navigationMapView(_ mapView: NavigationMapView, shapeFor routes: [Route]) -> MLNShape? {
```
Patrick-Kladek, line 568: I've annotated the whole class with `@objcMemebers`, so we don't need to mark every function with `@objc`

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -557,11 +508,13 @@ extension RouteMapViewController: NavigationViewDelegate {
 
     func navigationMapViewDidStartTrackingCourse(_ mapView: NavigationMapView) {
         self.navigationView.resumeButton.isHidden = true
+        self.navigationView.resumeButton.alpha = 0
```
Patrick-Kladek, line 511: adjusting alpha is needed to make animations looks good, here it would get out of sync, so need to adjust it as well

MapboxNavigation/StepsViewController.swift
```diff
@@ -112,40 +112,40 @@ public class StepsViewController: UIViewController {
     }
 
     func setupViews() {
-        view.translatesAutoresizingMaskIntoConstraints = false
+        self.view.translatesAutoresizingMaskIntoConstraints = false
```
Patrick-Kladek, line 115: only style changes, like added self

MapboxNavigation/UIView.swift
```diff
@@ -89,53 +89,6 @@ extension UIView {
         return view
     }
     
-    var safeArea: UIEdgeInsets {
-        guard #available(iOS 11.0, *) else { return .zero }
-        return safeAreaInsets
-    }
```
Patrick-Kladek, line 95: min deployment target is iOS12, so no need for this code

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -8,35 +8,58 @@ import UIKit
...
     var destination: Waypoint?
+    var showsEndOfRoute: Bool = true
+    var arrowCurrentStep: RouteStep?
+	
+    var contentInsets: UIEdgeInsets {
+        let top = self.navigationView.instructionsBannerContentView.bounds.height
+        let bottom = self.navigationView.bottomBannerView.bounds.height
+        return UIEdgeInsets(top: top, left: 0, bottom: bottom, right: 0)
+    }
+
```
Patrick-Kladek, line 62: in this file a lot of code was moved around

hactar (changes requested):
@Patrick-Kladek could you update the Readme.md section please? 

The Sample Code section is now not up to date:

> A demo app is currently not available. Please check the Mapbox repository or documentation for examples, especially on the forked version. You can try the provided demo app, which you need to first run carthage update --platform iOS --use-xcframeworks for in the root of this project.

Some help getting the example to run would be awesome.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -166,78 +197,16 @@ class RouteMapViewController: UIViewController {
...
+        self.currentLegIndexMapped = self.routeController?.routeProgress.legIndex ?? 0
+        self.currentStepIndexMapped = self.routeController?.routeProgress.currentLegProgress.stepIndex ?? 0
     }
 
     override func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
         self.removeTimer()
     }
 
-    func resumeNotifications() {
```
michaelkirk, line 179: Did anything change with this block or was this simply moved?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -248,19 +217,28 @@ class RouteMapViewController: UIViewController {
         self.mapView.setContentInset(self.contentInsets, animated: true, completionHandler: nil)
         self.mapView.setNeedsUpdateConstraints()
     }
+	
+    // MARK: - UIContentContainer
+	
+    override func preferredContentSizeDidChange(forChildContentContainer container: UIContentContainer) {
```
michaelkirk, line 223: Is this a new implementation or was it just moved?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -248,19 +217,28 @@ class RouteMapViewController: UIViewController {
...
+        UIView.animate(withDuration: 0.3, animations: self.view.layoutIfNeeded)
+    }
 
+    // MARK: - RouteMapViewController
+	
     func notifyDidReroute(route: Route) {
-        updateETA()
+        self.updateETA()
         self.currentStepIndexMapped = 0
+        guard let routeController else { return }
```
michaelkirk, line 232: I think a guard statement like this is good, and preferable to having `routeController?` all throughout this method.

I was expecting it at the top though, or is there a reason we want to call `self.updateETA()` and mutate `currentStepIndexMapped` even when routeController is nil? 

As well as avoiding the work, I think it communicates more clearly "this method should only be called while actively routing"

Also, since this is a *re-route* do we expect routeController ever would actually be nil? If not, can you add an `assertionFailure` to the guard statement's else clause?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -304,32 +267,13 @@ class RouteMapViewController: UIViewController {
...
-
-    @objc func updateInstructionsBanner(notification: NSNotification) {
-        guard let routeProgress = notification.userInfo?[RouteControllerNotificationUserInfoKey.routeProgressKey] as? RouteProgress else { return }
-        self.instructionsBannerView.update(for: routeProgress.currentLegProgress.currentStepProgress.currentVisualInstruction)
-        self.lanesView.update(for: routeProgress.currentLegProgress.currentStepProgress.currentVisualInstruction)
-        self.nextBannerView.update(for: routeProgress.currentLegProgress.currentStepProgress.currentVisualInstruction)
-    }
-
     func updateMapOverlays(for routeProgress: RouteProgress) {
+        guard let routeController else { return }
```
michaelkirk, line 271: If this should never happen, how about an assertionFailure?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -358,25 +302,6 @@ class RouteMapViewController: UIViewController {
         }
     }
 
-    private func showStatus(title: String, withSpinner spin: Bool = false, for time: TimeInterval, animated: Bool = true, interactive: Bool = false) {
```
michaelkirk, line 361: Did anything change here or was this simply moved?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -395,124 +320,148 @@ class RouteMapViewController: UIViewController {
             self.currentStepIndexMapped = routeProgress.currentLegProgress.stepIndex
         }
 
-        if self.annotatesSpokenInstructions {
-            self.mapView.showVoiceInstructionsOnMap(route: self.routeController.routeProgress.route)
+        if self.annotatesSpokenInstructions, let routeController {
```
michaelkirk, line 323: If `notifyDidChange(routeProgress: RouteProgress` should only ever be called when routeControler, how about a guard at the beginning of the method to make that clear.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -395,124 +320,148 @@ class RouteMapViewController: UIViewController {
...
-        endOfRoute.didMove(toParent: self)
+    /**
+     Updates the current road name label to reflect the road on which the user is currently traveling.
 
-        endOfRoute.dismissHandler = { [weak self] _, _ in
-            self?.routeController.endNavigation()
-            self?.delegate?.mapViewControllerDidDismiss(self!, byCanceling: false)
+     - parameter location: The user’s current location.
+     */
+    func labelCurrentRoad(at rawLocation: CLLocation, for snappedLoction: CLLocation? = nil) {
```
michaelkirk, line 333: Did anything change here or was this simply moved?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -395,124 +320,148 @@ class RouteMapViewController: UIViewController {
...
-        let animate = {
-            self.view.layoutIfNeeded()
-            self.navigationView.floatingStackView.alpha = 0.0
+        if let labelRoadNameCompletionHandler {
+            labelRoadNameCompletionHandler(true)
         }
+    }
 
-        let noAnimation = { animate(); completion?(true) }
+    func labelCurrentRoadFeature(at location: CLLocation) {
```
michaelkirk, line 356: Did anything change here or was this simply moved?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -395,124 +320,148 @@ class RouteMapViewController: UIViewController {
...
+            self.navigationView.wayNameView.isHidden = true
         }
-
-        guard duration > 0.0 else { return noAnimation() }
-        UIView.animate(withDuration: duration, delay: 0.0, options: [.curveLinear], animations: animate, completion: complete)
     }
+	
+    // MARK: End Of Route
+	
+    func transitionToEndNavigation(with duration: TimeInterval, completion: ((Bool) -> Void)? = nil) {
```
michaelkirk, line 424: Did anything change here or was this simply moved?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -395,124 +320,148 @@ class RouteMapViewController: UIViewController {
...
+
+        guard duration > 0.0 else { return noAnimation() }
+
+        self.navigationView.mapView.tracksUserCourse = false
+        UIView.animate(withDuration: duration, delay: 0.0, options: [.curveLinear], animations: animate, completion: completion)
     }
-}
 
-// MARK: - UIContentContainer
+    func hideEndOfRoute(duration: TimeInterval = 0.3, completion: ((Bool) -> Void)? = nil) {
```
michaelkirk, line 445: Did anything change here or was this simply moved?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -575,31 +528,6 @@ extension RouteMapViewController: NavigationViewDelegate {
     func didTapInstructionsBanner(_ sender: BaseInstructionsBannerView) {
         self.displayPreviewInstructions()
     }
-    
-    private func displayPreviewInstructions() {
```
michaelkirk, line 579: Did anything change here, or was this simply moved?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -644,110 +572,234 @@ extension RouteMapViewController: NavigationViewDelegate {
...
+
+            controller.didMove(toParent: self)
+            controller.dropDownAnimation()
+
+            self.stepsViewController = controller
+            return
+        }
+    }
+	
+    func roadFeature(for line: MLNPolylineFeature) -> (roadName: String?, shieldName: NSAttributedString?) {
```
michaelkirk, line 802: Is there a reason this method is no longer private?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -800,135 +852,29 @@ extension RouteMapViewController: NavigationViewDelegate {
...
 
         if self.annotatesSpokenInstructions {
-            self.mapView.showVoiceInstructionsOnMap(route: self.routeController.routeProgress.route)
-        }
-    }
-}
-
-// MARK: StepsViewControllerDelegate
-
-extension RouteMapViewController: StepsViewControllerDelegate {
```
michaelkirk, line 831: Did these change or just moved?

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -945,19 +891,3 @@ private extension UIView.AnimationOptions {
         }
     }
 }
-
-@objc protocol RouteMapViewControllerDelegate: NavigationMapViewDelegate, MLNMapViewDelegate, VisualInstructionDelegate {
```
michaelkirk, line 949: Did this change or just moved?

MapboxNavigationTests/Sources/Tests/NavigationViewControllerTests.swift
```diff
@@ -181,13 +183,14 @@ class NavigationViewControllerTests: XCTestCase {
 
         // wait for the style to load -- routes won't show without it.
         wait(for: [styleLoaded], timeout: 5)
-        navigationViewController.route = self.initialRoute
+//        navigationViewController.route = self.initialRoute
```
michaelkirk, line 186: nit: delete this?

MapboxNavigationTests/Sources/Tests/NavigationViewControllerTests.swift
```diff
@@ -257,7 +260,7 @@ class NavigationViewControllerTestable: NavigationViewController {
                   dayStyle: Style,
                   styleLoaded: XCTestExpectation) {
         self.styleLoadedExpectation = styleLoaded
-        super.init(for: route, dayStyle: dayStyle, directions: Directions(accessToken: "abc", host: ""), voiceController: FakeVoiceController())
+        super.init(dayStyle: dayStyle, directions: Directions(accessToken: "abc", host: ""), voiceController: FakeVoiceController())
```
michaelkirk, line 263: Maybe it's not necessary for the tests to pass currently, but I think you might save someone surprise in the future if you set the route they've passed in as an arg.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -644,110 +572,234 @@ extension RouteMapViewController: NavigationViewDelegate {
...
-            return CGPoint(x: mapView.bounds.midX, y: mapView.bounds.height * 0.4)
+        self.delegate?.navigationMapViewUserAnchorPoint?(mapView) ?? .zero
+    }
+}
+
+// MARK: StepsViewControllerDelegate
+
+extension RouteMapViewController: StepsViewControllerDelegate {
+    func stepsViewController(_ viewController: StepsViewController, didSelect legIndex: Int, stepIndex: Int, cell: StepTableViewCell) {
+        guard let routeController else { return }
```
michaelkirk, line 583: assertionFailure?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -18,6 +18,13 @@ public protocol NavigationViewControllerDelegate: VisualInstructionDelegate {
      - parameter canceled: True if the user dismissed the navigation view controller by tapping the Cancel button; false if the navigation view controller dismissed by some other means.
      */
     @objc optional func navigationViewControllerDidDismiss(_ navigationViewController: NavigationViewController, byCanceling canceled: Bool)
```
michaelkirk, line 20: Is this method called? I think it's been replaced by `navigationViewControllerDidFinish`, in which case we should delete it.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -18,6 +18,13 @@ public protocol NavigationViewControllerDelegate: VisualInstructionDelegate {
...
+	
+    /**
+     Called when user arrived at the destination of the trip.
+	 
+     - parameter navigationViewController: The navigation view controller that finished navigation.
+     */
+    @objc optional func navigationViewControllerDidFinish(_ navigationViewController: NavigationViewController)
```
michaelkirk, line 27:
```suggestion
    @objc optional func navigationViewControllerDidFinishRouting(_ navigationViewController: NavigationViewController)
```

Take it or leave it naming nit.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -190,61 +197,81 @@ public protocol NavigationViewControllerDelegate: VisualInstructionDelegate {
  
  - seealso: CarPlayNavigationViewController
  */
+@objcMembers
```
michaelkirk, line 200: I never knew about `@objcMembers`! Is the benefit that it saves you some keyboard typing?

Does it preclude us from using swift-only types in private methods?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -300,49 +308,30 @@ open class NavigationViewController: UIViewController {
...
-        self.currentStatusBarStyle
-    }
+    public var annotatesSpokenInstructions = false
     
+    // MARK: - Lifecycle
+	
     public required init?(coder aDecoder: NSCoder) {
+        self.directions = .shared
+        self.mapViewController = RouteMapViewController(routeController: self.routeController)
+        self.locationManager = NavigationLocationManager()
```
michaelkirk, line 332: Just so I understand: These assignments are a consequence of getting rid of the implicitly unwrapped optionals above, right?

(seems like a good tradeoff 👍)

MapboxNavigation/NavigationViewController.swift
michaelkirk: > Right now we offer 3 or 4 different init options.

Lets break this down a bit.

init 1: initWithCoder.

I literally don't know if this is usable. I think we just have it to satisfy the compiler. If it's not intended to be usable, we'd be better off replacing it with a useful `fatalError("use foo instead")`.

init 2: `@objc(initWithRoute:directions:styles:routeController:locationManager:voiceController:)` was recently deprecated (#45) because it relies on an implicit Style/tileserver (formerly this hit mapbox, now maplibre demo). It is useful to keep around because it will provide users with the upgrade path to the new init method.

init 3: Accepts a `Style`. I think we have to keep this one as it's the most general.

init 4: Most people don't care about the particulars of `Style`, and just want to give a style URL. I think it's the one most people will actually want to use.

So in summary, yes it's 4 inits, but I think we should keep them all. 

I think we should delete the deprecated method after it's been released long enough for the deprecation message to be useful. I won't be super mad though if you feel strongly about deleting it before the deprecation has been released.

I feel more strongly that we should keep both the `Style` and `styleURL` init flavors.

Disclaimer: I am biased because I deprecated init 2 and wrote init 3 + 4. 😉

MapboxNavigation/NavigationView.swift
```diff
@@ -163,18 +148,75 @@ open class NavigationView: UIView {
         self.commonInit()
     }
     
+    // MARK: - NavigationView
+    
+    override open func prepareForInterfaceBuilder() {
```
michaelkirk, line 153: Did any of this block change or was it just moved?

michaelkirk (commented):
Thanks for working on this!

I haven't looked at the Example app, but looked at the rest of the code. I'll check out the example app later, but first I want to try integrating this into my own app later today.

Sorry for the very many "did this change or was it moved" comments. It's challenging to review the mixture of functional changes and code reorganization in a single diff.

michaelkirk:
Testing on my own app, the changes required were minimal:
```diff
diff --git a/maps.earth/Directions/NavigationViewController.swift b/maps.earth/Directions/NavigationViewController.swift
index 57bca7e..7a62d32 100644
--- a/maps.earth/Directions/NavigationViewController.swift
+++ b/maps.earth/Directions/NavigationViewController.swift
@@ -51,16 +51,16 @@ struct MENavigationViewController: UIViewControllerRepresentable {
       simulatedLocationManager = nil
     }
     let vc = MapboxNavigation.NavigationViewController(
+      for: route,
       dayStyleURL: AppConfig().tileserverStyleUrl,
       directions: mlnDirections,
       locationManager: simulatedLocationManager
     )
-    vc.startNavigation(with: route)
     assert(vc.delegate == nil)
     vc.delegate = context.coordinator
     return vc
   }
```

It's nice to not break APIs, or deprecate them to guide users to the new method, but I don't feel that strongly about it in this case, as long as we provide migration instructions in the change log.

michaelkirk:
SimulatedLocationManger no longer seems to function. Do I need to do more than the aforementioned diff (https://github.com/maplibre/maplibre-navigation-ios/pull/54#issuecomment-2153238133) to migrate?

## With this branch

* location doesn't change.
* status banner is empty

https://github.com/maplibre/maplibre-navigation-ios/assets/217057/b8f1d936-5884-4aad-af9e-5142af3672e5

Note that the zooming/panning that starts at 5.5s is being done manually, just to show more clearly that the puck isn't moving.

## Before

*  location updates start firing immediately
* status banner shows my simulation speed

https://github.com/maplibre/maplibre-navigation-ios/assets/217057/7a9c2cf1-1273-4ce7-80cf-fa28dfd8f94b

MapboxNavigation/NavigationViewController.swift
michaelkirk: Ah, I missed https://github.com/maplibre/maplibre-navigation-ios/pull/54#discussion_r1629649939 - seems like it's primarily about saving some keyboard typing. Fine by me.

MapboxNavigation/NavigationViewController.swift
michaelkirk: I'd advocate for including this change in the breaking change log as well, it momentarily tripped me up when the critical logic in my `navigationViewControllerDidDismiss` implementation wasn't being run.

Example/Example.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -0,0 +1,59 @@
+{
+  "pins" : [
+    {
+      "identity" : "mapbox-directions-swift",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/flitsmeister/mapbox-directions-swift",
```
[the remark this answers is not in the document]
Patrick-Kladek: No, its just using this package as is

MapboxNavigation/NavigationViewController.swift
Patrick-Kladek: I'm a bit unhappy with the first init (deprecated) as we need to add backwards compatibility and that means double testing and special workarounds.

New way: `NavigationViewController.init(dayStyleURL: URL)`, add it as rootViewController and call `startNavigation()` when you are ready.

Old way: `NavigationViewController.init(for: route)` and this starts a navigation right away, user is expected to call `present(navigationViewController, animated: true)`. I think this currently doesn't work correctly as I've removed viewWillAppear & viewWillDisappear methods. So keeping this old init around adds a lot of complexity to keep both ways around.

Given that this will be in a major version release I think it's acceptable to break things and as long as we mention this in the release notes it shouldn't cause issues. But maybe I'm missing something important here and it does make sense to add this extra complexity.

MapboxNavigation/BottomBannerViewLayout.swift
```diff
@@ -5,34 +5,34 @@ extension BottomBannerView {
         let timeRemainingLabel = TimeRemainingLabel()
         timeRemainingLabel.translatesAutoresizingMaskIntoConstraints = false
         timeRemainingLabel.font = .systemFont(ofSize: 28, weight: .medium)
-        addSubview(timeRemainingLabel)
+        self.addSubview(timeRemainingLabel)
```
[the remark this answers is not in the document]
Patrick-Kladek: As far as I know we haven't decided. We do use SwiftFormat but it's not specified. The reason I add self is that without you loose context, especially in GitHub, in Xcode its fine with syntax highlighting. I've encountered various places with explicit & implicit self, so I started adding it where I touched the code.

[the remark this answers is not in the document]
Patrick-Kladek: Wd can always reformat the whole project with one command when we agree on one style.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -134,29 +170,24 @@ class RouteMapViewController: UIViewController {
...
             self.mapView.updateCourseTracking(location: location, animated: false)
-        } else if let coordinates = routeController.routeProgress.currentLegProgress.currentStep.coordinates, let firstCoordinate = coordinates.first, coordinates.count > 1 {
+        } else if let coordinates = self.routeController?.routeProgress.currentLegProgress.currentStep.coordinates, let firstCoordinate = coordinates.first, coordinates.count > 1 {
             let secondCoordinate = coordinates[1]
             let course = firstCoordinate.direction(to: secondCoordinate)
-            let newLocation = CLLocation(coordinate: routeController.location?.coordinate ?? firstCoordinate, altitude: 0, horizontalAccuracy: 0, verticalAccuracy: 0, course: course, speed: 0, timestamp: Date())
+            let newLocation = CLLocation(coordinate: self.routeController?.location?.coordinate ?? firstCoordinate, altitude: 0, horizontalAccuracy: 0, verticalAccuracy: 0, course: course, speed: 0, timestamp: Date())
             self.mapView.updateCourseTracking(location: newLocation, animated: false)
         } else {
             self.mapView.setCamera(self.tiltedCamera, animated: false)
```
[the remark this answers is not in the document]
Patrick-Kladek: Good catch with the camera, removing this line actually fixed a bug in my app, as I've added a swiftUI wrapper I wasn't sure who introduced this bug. Thanks.

Regarding the guard, I've tried it but I don't think it makes much difference, however if you have a strong opinion about it I can add it, just let me know.

Bildschirmfoto 2024-06-07 um 15 29 53

MapboxNavigation/RouteMapViewController.swift
Patrick-Kladek: just moved to private section

MapboxNavigation/RouteMapViewController.swift
Patrick-Kladek: just moved

MapboxNavigation/RouteMapViewController.swift
Patrick-Kladek: just moved

MapboxNavigation/RouteMapViewController.swift
Patrick-Kladek: just moved

MapboxNavigation/RouteMapViewController.swift
Patrick-Kladek: just moved

MapboxNavigation/RouteMapViewController.swift
Patrick-Kladek: this is new

MapboxNavigation/RouteMapViewController.swift
Patrick-Kladek: just moved

MapboxNavigation/RouteMapViewController.swift
Patrick-Kladek: just moved

MapboxNavigation/RouteMapViewController.swift
Patrick-Kladek: its still private, just in private extension

MapboxNavigation/RouteMapViewController.swift
Patrick-Kladek: just moved

MapboxNavigation/NavigationViewController.swift
Patrick-Kladek: yes, the same is happening in the other init methods

MapboxNavigation/NavigationView.swift
Patrick-Kladek: just moved

MapboxNavigation/RouteMapViewController.swift
Patrick-Kladek: good idea, it makes no sense to run call this method outside of a navigation, so routeController is required

MapboxNavigationTests/Sources/Tests/NavigationViewControllerTests.swift
Patrick-Kladek: good catch, thanks

MapboxNavigation/NavigationViewController.swift
Patrick-Kladek: This one is different. There are now two slightly confusing delegates.

One is called when the user taps the end button in the navigation UI - `func navigationViewControllerDidDismiss(_ navigationViewController: NavigationViewController, byCanceling canceled: Bool)`

The second one is called when the user arrived at the destination and the navigation ended therefore.

Will mention this in the changelog

Patrick-Kladek:
@michaelkirk I see you are trying to use this in SwiftUI right, I do have a working prototype in my App that is completely written in SwiftUI. Checkout https://github.com/stadiamaps/maplibre-swiftui-dsl-playground. Once this is merged I'll open a PR there.

Regarding your specific use case, you need to pass in a locationManager, otherwise it uses the real location of the device/simulator. Its done the same way in the example app.

```swift
let simulatedLocationManager = SimulatedLocationManager(route: route)
simulatedLocationManager.speedMultiplier = 2

self.viewController.startNavigation(with: route, locationManager: simulatedLocationManager)
```

MapboxNavigation/NavigationViewController.swift
```diff
@@ -389,191 +379,138 @@ open class NavigationViewController: UIViewController {
...
         self.resumeNotifications()
-        view.clipsToBounds = true
+        self.view.clipsToBounds = true
     }
     
-    override open func viewWillAppear(_ animated: Bool) {
-        super.viewWillAppear(animated)
+    // MARK: - NavigationViewController
+	
+    public func startNavigation(with route: Route, routeController: RouteController? = nil, locationManager: NavigationLocationManager = NavigationLocationManager()) {
```
michaelkirk, line 464: This routeController parameter is not currently being used. Did you intend to assign it?

And if so, does it even make sense to have a routeController parameter for init?

MapboxNavigation/BottomBannerViewLayout.swift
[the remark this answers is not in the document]
michaelkirk: > We can always reformat the whole project with one command when we agree on one style.

Sounds good. I'd support applying a big all at once formatting change independent of any functional change.

Applying stylistic preferences to functionally unchanged code makes the diff harder to review. So in the meanwhile, let's try to show some restraint.

MapboxNavigation/RouteMapViewController.swift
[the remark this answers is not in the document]
michaelkirk: I prefer the guard. Could it go at the top of the method? 

I don't feel that strongly about this though if you'd like to keep the optionals.

MapboxNavigation/NavigationViewController.swift
michaelkirk: Sorry to be obtuse, but can you point me to the line of code where `navigationViewControllerDidDismiss` is called?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -300,49 +308,30 @@ open class NavigationViewController: UIViewController {
...
-            self.mapViewController?.isUsedInConjunctionWithCarPlayWindow = self.isUsedInConjunctionWithCarPlayWindow
-        }
-    }
-    
-    var isConnectedToCarPlay: Bool {
-        if #available(iOS 12.0, *) {
-            CarPlayManager.shared.isConnectedToCarPlay
-        } else {
-            false
+            self.mapViewController.isUsedInConjunctionWithCarPlayWindow = self.isUsedInConjunctionWithCarPlayWindow
```
michaelkirk, line 318: Now that you've made `mapViewController` non-optional, how about making this var proxy to the central underlying var, rather than copying the state around.

It seems like it would be easier to keep things in sync, e.g. if `mapViewController` edits `isUsedInConjunctionWithCarPlayWindow` directly or something.

It's a minor thing, so take it or leave it.

Example/example/SceneDelegate.swift
```diff
@@ -0,0 +1,127 @@
...
+            self.viewController.startNavigation(with: route, locationManager: simulatedLocationManager)
+        }
+    }
+    
+    @objc
+    func cameraButtonTapped() {
+        guard let waypoint = self.waypoints.randomElement() else { return }
+        
+        func randomCLLocationDistance(min: CLLocationDistance, max: CLLocationDistance) -> CLLocationDistance {
+            CLLocationDistance(arc4random_uniform(UInt32(max - min)) + UInt32(min))
```
michaelkirk, line 106: Might as well address these lints for new code.

MapboxNavigation/NavigationViewController.swift
Patrick-Kladek: You were right from the beginning, this should be removed, only the car play related methods need to stay

Example/example/SceneDelegate.swift
Patrick-Kladek: Sadly SwiftLint is not running in Xcode, only on CI so I was not aware of those warnings

MapboxNavigation/UIView.swift
boldtrn: Yes makes perfect sense. If variables are removed or if parameters of public methods are changed, it would be great if this would be mentioned in the changelog :)

Patrick-Kladek:
@boldtrn updated the changelog & readme with migration instructions

CHANGELOG.md
```diff
@@ -24,6 +25,21 @@
...
+- Start & Stop Navigation in existing Map
+	- Removed: `NavigationViewController(for route: Route, dayStyle: Style, routeController: RouteController? = nil, locationManager: NavigationLocationManager? = nil, voiceController: RouteVoiceController? = nil)` use `NavigationViewController(dayStyleURL: URL, nightStyleURL: URL? = nil,directions: Directions = .shared, voiceController: RouteVoiceController = RouteVoiceController())` followed by `startNavigation(with route: Route)` instead.
+	- To simulate a route, pass a `SimulatedLocationManager` to `startNavigation()` function: 
+
+	```swift
+	let vc = NavigationViewController(dayStyleURL: AppConfig().tileserverStyleUrl)
+	
+	if Env.current.simulateLocationForTesting {
+	    let simulatedLocationManager = SimulatedLocationManager(route: route)
+        simulatedLocationManager.speedMultiplier = 5
```
michaelkirk, line 37: nit: weird indent

README.md
```diff
@@ -18,6 +18,49 @@ All issues are covered with this SDK.
 - Transitioned from the [Mapbox SDK](https://github.com/mapbox/mapbox-gl-native-ios) (version 4.3) to [Maplibre Maps SDK](https://github.com/maplibre/maplibre-gl-native) (version 6.0.0)
 - Added optional config parameter in NavigationMapView constructor to customize certain properties like route line color
 
+# Migrating from v2 to v3
+
+Maplibre v3 allows you to start a navigation in an existing Map, so no modal ViewController needs to be presented over an existing map as before. This results in some breaking changes.
```
michaelkirk, line 23:
```suggestion
MaplibreNavigation v3 allows you to start a navigation in an existing Map, so no modal ViewController needs to be presented over an existing map as before. This results in some breaking changes. 
```

README.md
```diff
@@ -18,6 +18,49 @@ All issues are covered with this SDK.
 - Transitioned from the [Mapbox SDK](https://github.com/mapbox/mapbox-gl-native-ios) (version 4.3) to [Maplibre Maps SDK](https://github.com/maplibre/maplibre-gl-native) (version 6.0.0)
 - Added optional config parameter in NavigationMapView constructor to customize certain properties like route line color
 
+# Migrating from v2 to v3
+
+Maplibre v3 allows you to start a navigation in an existing Map, so no modal ViewController needs to be presented over an existing map as before. This results in some breaking changes. 
+
+### Step 1:
+
+Replace your ViewController that hosts the mapView with a `NavigationViewController`. We suggest to create a subclass of `NavigationViewController` and override init & call `super.init(dayStyle:)` or `super.init(dayStyleURL:)`. NavigationViewController will not do anything on its own until you start a navigation.
```
michaelkirk, line 27:
```suggestion
Replace your ViewController that hosts the mapView with a `NavigationViewController`. We suggest to create a subclass of `NavigationViewController` and override init & call `super.init(dayStyle:)` or `super.init(dayStyleURL:)`. With v3, NavigationViewController will not start navigation until you request it.
```

README.md
```diff
@@ -18,6 +18,49 @@ All issues are covered with this SDK.
...
+Start the navigation by calling `startNavigation(with: route)`. If you want to simulate your route, you need to pass in the optional locationManager parameter, otherwise the real location of the device will be used.
+
+```swift
+func locationManager(for route: Route) -> NavigationLocationManager {
+#if targetEnvironment(simulator)
+	let locationManager = SimulatedLocationManager(route: route)
+	locationManager.speedMultiplier = 2
+	return locationManager
+#else
+	return NavigationLocationManager()
```
michaelkirk, line 40:
```
#if targetEnvironment(simulator) 
    let locationManager = SimulatedLocationManager(route: route)
    locationManager.speedMultiplier = 2
    self.startNavigation(with: route, locationManager: locationManager)
#else
    self.startNavigation(with: route)
#endif
```

Slightly shorter and we can avoid the concept of `NavigationLocationManager()` - people can just use the default value.

README.md
```diff
@@ -18,6 +18,49 @@ All issues are covered with this SDK.
...
+```
+
+### Step 3:
+
+Make your ViewController conform to `NavigationViewControllerDelegate` and implement `navigationViewControllerDidFinishRouting(_ navigationViewController: NavigationViewController)` to transition back to normal map mode. This delegate is called when the user arrives at the destination or cancels the navigation.
+
+```swift
+extension SceneDelegate: NavigationViewControllerDelegate {
+    func navigationViewControllerDidFinishRouting(_ navigationViewController: NavigationViewController) {
+        navigationViewController.endNavigation()
```
michaelkirk, line 54: `endNavigation` is a new method right?

Is there a reason `navigationViewController` doesn't call `endNavigation` itself, rather than requiring implementors to have a specific implementation for their delegate?

README.md
```diff
@@ -42,114 +85,15 @@ Install this package using the [Swift Package Manager](https://www.swift.org/doc
 - **Have a bug to report?** [Open an issue](https://github.com/maplibre/maplibre-navigation-ios/issues). If possible, include the version of Maplibre Services, a full log, and a project that shows the issue.
 - **Have a feature request?** [Open an issue](https://github.com/maplibre/maplibre-navigation-ios/issues/new). Tell us what the feature should do and why you want the feature.
 
-## <a name="sample-code">Sample code
+## Sample code
 
-A demo app is currently not available. Please check the Mapbox repository or documentation for examples, especially on the forked version. You can try the provided demo app, which you need to first run `carthage update --platform iOS --use-xcframeworks` for in the root of this project.
+We do provide a limited example app but its not functional right out of the box.
```
michaelkirk, line 90: 🎉

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changes to the Mapbox Navigation SDK for iOS
 
 ## Unreleased
+ - Map can be used without an active navigation, navigation can be started anytime with `startNavigation(route: Route)`.
```
michaelkirk, line 4: Did you intend to have two changelog entries? This one seems redundant with the one you added at the end. Can you consolidate them?

README.md
```diff
@@ -162,3 +106,4 @@ Code is [licensed](LICENSE.md) under MIT and ISC.
 ISC is meant to be functionally equivalent to the MIT license.
 
 Copyright (c) 2022 MapLibre contributors
+®®®
```
michaelkirk, line 109: Delete this line?

michaelkirk:
I'm running into a challenge trying to integrate this into my app. It's not necessarily a problem with this PR, but I wanted to get your thoughts @Patrick-Kladek.

For context, my app (https://apps.apple.com/us/app/maps-earth/id6479329515)  is a general purpose mapping app like Google Maps or Apple Maps. You can search for places, see routes between them rendered on the map, and (as of recently!) navigate one of those routes using this repository.

To achieve this, I have a SwiftUI `UIViewRepresentable` with  `UIViewType = MLNMapView` which I use for everything except navigation. For navigation, I modally present a new map via `MapboxNavigation.NavigationViewController`.

This works, but like you, I'd like to consolidate these two maps and navigate from the same map that I use for searching and everything else in the app. I can almost do that with this PR, but it breaks some other things.

This is because my previous approach leverages `MLNMapViewDelegate` methods. With this PR, I can't assign my existing delegate to  `navigationViewController.mapView.delegate` because it is already set to `MapboxNavigation.RouteMapViewController`.

Did you run into this @Patrick-Kladek? How did you address it?

MapboxNavigation/UIView.swift
ianthetechie: > Any type members added in an extension have the same default access level as type members declared in the original type being extended. If you extend a public or internal type, any new type members you add have a default access level of internal. 

The extension doesn't have an access modifier on it, and none of these removed computed props had them either, so I think this should be safe to remove (all were internal and thus not part of the public interface).

README.md
Patrick-Kladek: I thought this gives the most flexibility. Two examples:
* when you want to display an arrived view.
* If you want to support multi stop routing, you could then start the next navigation without animating back and forth

ianthetechie: It looks like it does call `endNavigation()` at least when I'm reading in right now.

README.md
```diff
@@ -18,6 +18,49 @@ All issues are covered with this SDK.
...
+extension SceneDelegate: NavigationViewControllerDelegate {
+    func navigationViewControllerDidFinishRouting(_ navigationViewController: NavigationViewController) {
+        navigationViewController.endNavigation()
+    }
+}
+```
+
+### Backwards compatibility
+
+If for some reason, you want to keep the old way of presenting a navigation modally, you can still do that. Simply call `startNavigation(with: route)` right after creating the `NavigationViewController`.
```
ianthetechie, line 57:
```suggestion
If you want to keep the old way of presenting a navigation modally, you can still do that. Simply call `startNavigation(with: route)` right after creating the `NavigationViewController`.
```

README.md
ianthetechie: This is great!

Maybe you could elaborate slightly on what exactly it does / does not do in the README.

README.md
```diff
@@ -18,6 +18,49 @@ All issues are covered with this SDK.
...
+	return NavigationLocationManager()
+#endif
+}
+
+self.startNavigation(with: route, locationManager: locationManager(for : route))
+```
+
+### Step 3:
+
+Make your ViewController conform to `NavigationViewControllerDelegate` and implement `navigationViewControllerDidFinishRouting(_ navigationViewController: NavigationViewController)` to transition back to normal map mode. This delegate is called when the user arrives at the destination or cancels the navigation.
```
ianthetechie, line 45: It's probably worth noting (not sure if here or elsewhere?) that the _similar_ method this replaces (dismissal; maybe via cancellation) has also changed semantics slightly. Internal logic used to depend on whether the method was implemented or not. This is no longer the case. I like the new version better, but think we should probably document this subtlety.

Patrick-Kladek:
@michaelkirk first of all cool app.

What delegate methods do you need exactly and for what? I'm asking because thats a good point you've raised and could become a problem for us as well. 

We started a new app from scratch in SwiftUI and found [maplibre-swiftui-dsl-playground](https://github.com/stadiamaps/maplibre-swiftui-dsl-playground). It is managing the camera and additional layers for us. This PR adds navigation: https://github.com/stadiamaps/maplibre-swiftui-dsl-playground/pull/39

Now that you've raised this, I've realized it assigns the MapView.delegate to itself - interestingly I didn't notice it and everything is working but that doesn't mean we'll keep it like that.

I think for this use case NSProxy could become useful. I've never used it before but will try it and report my findings. If that doesn't work then I afraid we would need to manually redirect the delegates to `NavigationViewControllerDelegate`.

ianthetechie:
Looks good to me in principle. I'll be honest it was a slog reading all of the non-functional changes 😅

michaelkirk:
> What delegate methods do you need exactly and for what? I'm asking because thats a good point you've raised and could become a problem for us as well.

Currently I'm using:
```
// add/modify layers loaded from URL
func mapView(_ mapView: MLNMapView, didFinishLoading style: MLNStyle)

// Searching puts a bunch of markers on the map. Tap a specific marker to "select" it and see details
func mapView(_ mapView: MLNMapView, didSelect annotation: MLNAnnotation)

// Certain markers are drawn differently
func mapView(_ mapView: MLNMapView, viewFor annotation: MLNAnnotation) -> MLNAnnotationView?

// I render a custom control for MLNUserTrackingMode
func mapView(_ mapView: MLNMapView, didChange mode: MLNUserTrackingMode, animated: Bool)
```

But it stands to reason that I'd want to use many of them.

> NSProxy

I haven't thought about it too hard, but NSProxy seems like an interesting avenue! Good idea.

michaelkirk:
Is anyone able to run the example app?

```
open Example.xcodeproj
```

When I try to build I get:
> Missing package project 'MapboxNavigation'

Screenshot 2024-06-12 at 16 25 40

I think relatedly, I'm not seeing any of the dependencies in the left hand file explorer:

Screenshot 2024-06-12 at 16 25 50

I was able to resolve it locally by deleting and re-adding the dependencies:
image

I can see that you've already specified the deps - they're in the Package.swift.resolved - with mapslibre-navigation as a local/development dependency, which all seems like it should work. 

I'm not sure what's going on. 🤔 
I'm wondering if this is just me, or if other people experienced this.

README.md
Patrick-Kladek: This function `@objc optional func navigationViewControllerDidDismiss(_ navigationViewController: NavigationViewController, byCanceling canceled: Bool)` was removed. Will mention it in the README.

Patrick-Kladek:
@michaelkirk I guess you have the same package opened in another window, SPM will refuse to open the same package twice, so in this case make sure you've closed all other Xcode windows. Then it works for me

michaelkirk:
> I guess you have the same package opened in another window, SPM will refuse to open the same package twice, so in this case make sure you've closed all other Xcode windows. Then it works for me 

That fixed it! Thank you.

michaelkirk:
https://github.com/maplibre/maplibre-navigation-ios/blob/main/MapboxNavigation/RouteMapViewController.swift#L66

```swift
class RouteMapViewController: UIViewController {
  ...
  weak var delegate: RouteMapViewControllerDelegate?
  ...
}

// RouteMapViewControllerDelegate has a few of its own methods, but it also extends 
// the delegate protocols used internally by RouteMapViewController. 
@objc protocol RouteMapViewControllerDelegate: NavigationMapViewDelegate, MLNMapViewDelegate, VisualInstructionDelegate {
  func mapViewControllerDidDismiss(_ mapViewController: RouteMapViewController, byCanceling canceled: Bool)
  func mapViewControllerShouldAnnotateSpokenInstructions(_ routeMapViewController: RouteMapViewController) -> Bool
  @objc func mapViewController(_ mapViewController: RouteMapViewController, roadNameAt location: CLLocation) -> String?
}

// Note: NavigationViewDelegate: NavigationMapViewDelegate, MLNMapViewDelegate
extension RouteMapViewController: NavigationViewDelegate {
  ....
  // most methods are simply forwarded on to the delegate the implementor has set
  func navigationMapView(_ mapView: MLNMapView, imageFor annotation: MLNAnnotation) -> MLNAnnotationImage? {
    self.delegate?.navigationMapView?(mapView, imageFor: annotation)
  }
  // most methods are simply forwarded on to the delegate the implementor has set
  func navigationMapView(_ mapView: MLNMapView, viewFor annotation: MLNAnnotation) -> MLNAnnotationView? {
    self.delegate?.navigationMapView?(mapView, viewFor: annotation)
  }
  ....
  // but methods RouteMapViewController needs to use internally do their own work before forwarding on to the delegate the implementor has set
  func navigationMapViewUserAnchorPoint(_ mapView: NavigationMapView) -> CGPoint {
    // If the end of route component is showing, then put the anchor point slightly above the middle of the map
    if self.navigationView.endOfRouteView != nil, let show = navigationView.endOfRouteShowConstraint, show.isActive {
        return CGPoint(x: mapView.bounds.midX, y: mapView.bounds.height * 0.4)
    }

    // otherwise, ask the delegate or return .zero
    return self.delegate?.navigationMapViewUserAnchorPoint?(mapView) ?? .zero
  }
}
```

I like this approach, and I think it's how I'd expect this class to work. What would you think of doing something similar for the NavigationViewController.delegate?

I am concerned that when implementing it, there's potentially a lot of boilerplate, making it easy to miss a method, especially if new methods are added to the delegate. But maybe your idea about using NSProxy could help mitigate that @Patrick-Kladek.

Patrick-Kladek:
I'm afraid the NSProxy approach will not work. Here is what I got:

Bildschirmfoto 2024-06-18 um 20 05 56

I've added a proxy class, that will intercept methods and then forward them to the actual receiver. This works fine, however it's not possible to send the same message to another receiver. In `forwardingTarget(for:)` I want to forward it to `additionalResponder`. My problem is I can't get the parameters, for this I would need `NSInvocation` and .... this is not available in swift 😤. If it was, one could use NSProxy, thats also the reason I've used NSObject as my base class.

The only way to achieve this, is by typing a lot of boilerplate code: https://forums.swift.org/t/nsproxy-dynamic-casting-to-swift-or-objc-class-behaves-differently/2909/4 or write this in objc which I don't want to but tried as well. The problem there is, I can't make the Proxy class conform to `RouteMapViewControllerDelegate` as its in another target.

michaelkirk:
> I'm afraid the NSProxy approach will not work. Here is what I got:

What do you think we should do @Patrick-Kladek? 

I think if it were me at this point I would just hammer out the boiler plate and pray. It's admittedly not a very satisfying solution.

Patrick-Kladek:
@michaelkirk I think it makes the most sense to just add the delegate methods needed.

For now I've added these 2, the other two are already part of `NavigationViewController` so your app shouldn't have an issue. If someone needs another delegate forwarded he can submit a PR otherwise we will never get this finished if we start thinking about every possible edge case someone might encounter.

Bildschirmfoto 2024-06-18 um 21 55 37

michaelkirk:
Just to make sure I understand your proposal:

`NavigationViewControllerDelegate` will not conform to `MLNMapViewDelegate`, but as needed we'll port similar method signatures from `MLNMapViewDelegate` (and others) to `NavigationViewControllerDelegate`.

Is that right? If so, that seems reasonable to me and less error prone than writing all the delegate methods and hoping they all stay in sync.

hactar:
@michaelkirk @Patrick-Kladek I'm following the "delegate" discussion and maybe I'm not fully understanding it, so let me sum up what I think this is about, and then add my thoughts. Please correct me if I'm wrong.

MapLibre has a few delegate functions that are commonly used by apps to customize MapLibre. MapLibre Navigation has its own delegate implementation for a few of these functions, and so blocks users implementing their own delegate here. Now you're discussion options on how to call both MapLibre Navigation delegate implementations and users implementations at the same time.

Assuming I read you correctly, then I'm not sure thats a feasible solution: implementation details in the dev's delegate and the maplibre navigation delegate could clash and cause the weirdest bugs.

For me, at this stage of this implementation of "in existing map", I'd see two states the MapView can be in: Navigating or not-navigating. In Navigating mode, MapLibre Navigation delegates are called, and only these are called, the user cannot override them (pre this PR they can't do so either while navigating in the sheet). But, in non navigation mode, the maplibre navigation delegates do not need to be called at all, and instead the user's delegate methods are to be called.

This of course does not unlock full customizability during navigation, but I'd argue this exceeds the scope of this ticket for now. All we'd need for this ticket is for MapLibre navigation, before it kicks of navigation mode, to check if the dev has set up a custom delegate for the mapview, if yes, store it in a var, replace the delegate with its own, and when navigation ends, restore the dev's delegate again.

Patrick-Kladek:
@hactar NavigationViewController has some delegate methods but not all that the underlying MLNMapView offers. That makes it tempting to change the maps delegate which will break navigation. One good example here is `didFinishLoadingStyle`. 

I don't think it makes sense to change the logic if an optional delegate is implemented, however the callbacks can be useful. 

My idea with NSProxy was to send the same function to two different listeners as it would be future proof. As this is not possible I've forwarded 2 methods without changing any logic

michaelkirk:
> MapLibre has a few delegate functions that are commonly used by apps to customize MapLibre. MapLibre Navigation has its own delegate implementation for a few of these functions, and so blocks users implementing their own delegate here. Now you're discussion options on how to call both MapLibre Navigation delegate implementations and a developer's implementations at the same time.

Thanks for restating. I think that accurately reflects my understanding of the problem. 👍 

> implementation details in the dev's delegate and the maplibre navigation delegate could clash and cause the weirdest bugs.

I agree that this functionality could be prone to conflicts and weird bugs. 

For a specific example, I have a hard time imagining a desirable outcome of calling a users implementation of:

```
- (BOOL)mapView:(MLNMapView *)mapView shouldChangeFromCamera:(MLNMapCamera *)oldCamera toCamera:(MLNMapCamera *)newCamera;
```

during navigation. But who knows?

>  When switching to navigation mode, the developer is willingly giving up control of the map: MapLibre Navigation delegates are called, and only these are called, the user cannot override them (pre this PR they can't do so either while navigating in the sheet)

I'm concerned introducing the concept of "navigating" and "not navigating" modes, and only calling the implementor's delegates methods when in "not navigating" mode could confuse the implementor.

e.g. I think as an implementor, if I had these methods implemented on my MLNMapViewDelegate, I'd expect them to be called even during navigation:
```
- (nullable MLNAnnotationView *)mapView:(MLNMapView *)mapView viewForAnnotation:(id <MLNAnnotation>)annotation;
- (CGFloat)mapView:(MLNMapView *)mapView alphaForShapeAnnotation:(MLNShape *)annotation;
- (void)mapViewRegionIsChanging:(MLNMapView *)mapView;
- (void)mapViewWillStartRenderingFrame:(MLNMapView *)mapView;
- (nullable UIImage *)mapView:(MLNMapView *)mapView didFailToLoadImage:(NSString *)imageName;
```

It's true that I could probably do something in those delegate methods that break the navigation experience. But ultimately if we want to support navigating from the same map we do everything else in, the software is going to get more complicated.

I think @Patrick-Kladek approach of "opting in" inner delegate methods to the `NavigationViewControllerDelegate` is a good middle ground. It's clear what hooks we support, we try to avoid the ones that are most likely to be problematic, and we can add more as the need presents itself.

hactar:
> I'm concerned introducing the concept of "navigating" and "not navigating" modes, and only calling the implementor's delegates methods when in "not navigating" mode could confuse the implementor.
> 
> e.g. I think as an implementor, if I had these methods implemented on my MLNMapViewDelegate, I'd expect them to be called even during navigation:
> 
> ```
> - (nullable MLNAnnotationView *)mapView:(MLNMapView *)mapView viewForAnnotation:(id <MLNAnnotation>)annotation;
> - (CGFloat)mapView:(MLNMapView *)mapView alphaForShapeAnnotation:(MLNShape *)annotation;
> - (void)mapViewRegionIsChanging:(MLNMapView *)mapView;
> - (void)mapViewWillStartRenderingFrame:(MLNMapView *)mapView;
> - (nullable UIImage *)mapView:(MLNMapView *)mapView didFailToLoadImage:(NSString *)imageName;
> ```
> 
> It's true that I could probably do something in those delegate methods that break the navigation experience. But ultimately if we want to support navigating from the same map we do everything else in, the software is going to get more complicated.
> 
> I think @Patrick-Kladek approach of "opting in" inner delegate methods to the `NavigationViewControllerDelegate` is a good middle ground. It's clear what hooks we support, we try to avoid the ones that are most likely to be problematic, and we can add more as the need presents itself.

While I don't agree that a developer would be "confuse[d]"  - the documentation could clearly state this with something like "While a `Route` is attached to your `NavigationViewController`, your `MLNMapViewDelegate` will not be called" - I do agree that your approach is more powerful and allows for more customization. It is more dangerous -  and I'd argue actually more likely to cause confusion than a clear "you're in charge/we're in charge". But if you two find a solution where both delegates can be called in a safe way, or can be called in an unsafe way, and then document that this is unsafe and the developer needs to ensure that their delegate implementation doesn't clash with MapLibre's, it wouldn't be a blocker for me 🤷‍♂️.

hactar (approved):
Gonna mark this as "approved" as someone has to be the first, but not merging until there's more people that approve. My argument for approval:
- This works
- We have done a 3.0 release already, so "main" should now reflect a development state for 4.0, and this is the big building block for 4.0.
- The sooner we merge this, the quicker any other work done is based on this, instead of conflicting with it
- Merging this allows other people to contribute improvements to it directly, instead of having to discuss this in the discussion here.

hactar:
@Patrick-Kladek Tests not passing though, so please take a look at that.

Patrick-Kladek:
Tests are passing now

Patrick-Kladek:
The plan was to get this merged by Monday but was delayed because of #58. Now the week nearly passed and I can't see any progress. Can we please get this merged today, I don't see any blockers or risks as we have a working 3.0 release.

boldtrn (approved):
@Patrick-Kladek, my understanding was that we first merge #58 and then #54. 

I am fine with merging this PR first, I haven't seen any issues when running the code or when reviewing it.

Patrick-Kladek:
@boldtrn thanks for approving. It's getting a bit frustrating for me as there is no progress in #58 in the last 2 days and thats why this here is on hold. But maybe should have poked in the other PR.

boldtrn:
No worries, I can understand. I am currently reviewing this repo with priority already. Reviews take their time if you do them seriously, so It's not always possible to do something on the same day. Feel free to merge this PR or wait on #58. I am fine with either way. 

Maybe @michaelkirk would like to chime in as well, I think he is in the US timezone, so maybe let's give him the opportunity to give this one more check as well?

hactar:
Alright, we have 3 approvals, I'm gonna do the honors. Thanks @Patrick-Kladek for the contributions and iterations. And thanks for all coding and reviewing contributions, I know this wasn't a small change ❤️

PR #58, opened 2024-06-13
Fixup EndOfRouteViewController

### LINKED ISSUE #57: crash when route ends: [Storyboard] Unknown class _TtC40maplibre-navigation-ios_MapboxNavigation13RatingControl in Interface Builder file.
### Steps to Trigger Behavior

1. Start navigation and wait for it to complete (e.g. with a simulatedLocationManger).

I don't have an example handy, but if you're interested I can put one together.

### Expected Behavior

The endOfRouteViewController should be presented

### Actual Behavior

The app crashes due to:

Screenshot 2024-06-12 at 17 42 55

> [Storyboard] Unknown class RatingControl in Interface Builder file.

### Version(s) affected

- Platform: iOS
- OS version: iOS17
- Device type: simulator (haven't tried on device)
- Emulator/ Simulator: yes
- Development OS: osx 14.5 (23F79)
- maplibre-navigation-ios: HEAD eaee92b3
- MapLibre GL version [e.g. 6.3.0]: 6.4.1

### Additional context

One thing that confused me is that there seems to be *another* bug whereby the end of route logic will *never* fire unless you are using spoken directions, which I hadn't been. I'll open up an issue for that momentarily. **(update - here it is:  https://github.com/maplibre/maplibre-navigation-ios/pull/58)**

### LINKED ISSUE #62: Feedback in EndOfRouteViewController goes nowhere.
## Motivation

The EndOfRouteViewController looks like this:
Screenshot 2024-06-13 at 16 55 03

If you click on one of the stars, a text field appears:
Screenshot 2024-06-13 at 16 55 09

There's currently a couple serious issues with the EndOfRouteViewController (EORVC as the git log affectionately calls it) #57, #58, (plus one more crash I haven't filed yet due to an invalid animation curve during keyboard presentation). 

I've work around them all, but then I finally thought to ask: What is this all for? 

I'm not currently collecting these end-of-trip feeback anywhere (though I'd actually like to at some point). I wondered if there was some sort of pluggable feedback service? It turns out there was a long time ago in the mapbox days, but it was removed back in: https://github.com/flitsmeister/maplibre-navigation-ios/pull/1 . These controls seem like a non-functioning remnant of that system.

## Design Alternatives

1. Leave the EORVC but delete the feedback elements. User sees only the "You have arrived" / "End Navigation" elements. (this is my preferred alternative)

2. ...or Delete the EORVC altogether. MaplibreNavigation would only fire a delegate hook, and the implementor is expected to end the route and provide any UI.

3. ...or Keep the EORVC  with its rating/feedback elements, but make it easy to hide them. Return the value of the feedback fields to the delegate method when the user taps "End Navigation" so that the implementor can handle it. This is the most functional, pre-supposes that the implementer wants to collect exactly this information. **Update: I realize now the EORVC already does return the rating/comment in the delegate callback... they're just ignored by `func embedEndOfRoute()`**. I still think we should drop it and say if people want to collect feedback or whatever they'll need to BYO EORVC.

I think I prefer alternative 1 - keeping the EORVC, but deleting the feedback elements and making it possible for the implementor to extend/replace this view.

I prefer this approach because I think it is coherent to present some kind of "You're Done!" state  arrives rather than leaving them in perpetual routing purgatory (and potentially re-reouting them if they subsequently wander).

I do think it's likely that people will want the ability to customize this UI (for example, I actually do want to collect feedback on my routing, but with different UI). I still think it's worthwhile to present some kind of basic UI. By presenting people with a default, that gets them a head start on a coherent navigation flow and if they do want a custom view there, they now have a way to trace back from what they already see in the UI to the specific place in the code they need to change rather than exhaustively cocvering such and such delegate documentation.

michaelkirk:
~~DO NOT MERGE: This currently exposes a crashes due to another issue: see https://github.com/maplibre/maplibre-navigation-ios/issues/57~~ Fixed with latest commit.

~~I think we'll eventually want to merge this and it might be helpful if you're trying to debug #57, which I'm currently stuck on.~~

Fixes #57 
Fixes #62 

---

Previously, having voice instructions was a requirement of ending the route.

Now we'll end the route when banner instructions are done *and* only consider voice instructions if there were any to begin with.

I also fell down the rabbit hole of fixing one bug after another once the EndOfRouteViewController (EORVC) was presented, otherwise, it's arguably better not to show it at all.

In particular, I also removed the "feedback" UI from the EORVC since the feedback mechanism was removed. See #62


MapboxCoreNavigation/RouteController.swift
```diff
@@ -424,19 +424,23 @@ extension RouteController: CLLocationManagerDelegate {
 
     func updateRouteLegProgress(for location: CLLocation) {
         let currentDestination = self.routeProgress.currentLeg.destination
-        guard let remainingVoiceInstructions = routeProgress.currentLegProgress.currentStepProgress.remainingSpokenInstructions else { return }
```
michaelkirk, line 427: This guard is what was throwing us off. We'd unconditionally return without calling `didArriveAt:` if not using spoken instructions.

michaelkirk:
I've fixed the crash, though as I've documented in my commit message, I don't really understand what's going on.

I don't know very much about storyboards nor swift package manager, but my guess is that there's some problem happening between the interaction of storyboards and SPM.

Here's what I've observed:

Before this change, here's the storyboard UI for the RatingControl element:

Screenshot 2024-06-13 at 13 36 19

I see this error when presenting the EndOfRouteViewController:

    [Storyboard] Unknown class _TtC40maplibre-navigation-ios_MapboxNavigation13RatingControl in Interface Builder file.

So the ViewController is successfully loaded from the bundle, but `RatingControl` has some prefix applied to the name, presumably that is some dynamic prefix thing that the compiler adds to namespace modules.

If I uncheck "Inherit Module from Target", the "Module" text field becomes empty:

Screenshot 2024-06-13 at 13 37 20

Re-running my route, upon completion of the route, it crashes at the same point, but with a slightly different error:  

    [Storyboard] Unknown class RatingControl in Interface Builder file.

Note that the RatingControl class is simple now, lacking the generated prefix from the first error.

If I then manually type "MapboxNavigation" into the "Module" field like this:

Screenshot 2024-06-13 at 13 36 09

The end of route controller now presents as expected. This is the change I'm proposing.

michaelkirk:
I just keep going down the rabbit hole here. 🐇

Marking as draft again until 
https://github.com/maplibre/maplibre-navigation-ios/issues/62 is resolved.

michaelkirk:
I think I've arrived at something worthwhile. 

Ultimately I've: 
- fixed the logic that was preventing the EORVC from presenting #58
- fixed the crash that occurred when the EORVC finally was presented #57
- fixed the crash that happened when you tried to interact with the feedback text field (I never opened an issue, but it's https://github.com/maplibre/maplibre-navigation-ios/pull/58/commits/2bf7799f7a0c356cb57039b84f2781fddbc3b931)
- fixed a layout bug with the keyboard animation: (I never opened an issue: https://github.com/maplibre/maplibre-navigation-ios/pull/58/commits/140236852c4807d71f0b5c61d54d3858eec58cc5)
- removed all the feedback fields per #62

I regret that it ballooned into something as big as it did, but I think it's kind of the smallest reasonably coherent set of work. I could upstream the above changes individually, or you might find reviewing the individual commits helpful.

I did take some liberty in my implementation in that I replaced the EORVC storyboard with programatic view construction. If this is viewed as distasteful, I can rework my changes to keep the storyboard in play.

My motivation for getting rid of the storyboard was:

1. I don't know very much about storyboards and they seem complicated and somewhat magical. For example, I really don't understand how I fixed the crash in #57.

2. It was the *only* storyboard view in the framework. It seems nice to reduce paradigms.


Screenshot 2024-06-13 at 19 20 50

Here's a waypoint with a pathologically long name so you can see how the layout grows:
Screenshot 2024-06-13 at 19 15 37

scripts/file_conversion.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env/bash
+#!/usr/bin/env bash
```
michaelkirk, line 1: typo?

scripts/extract_localizable.sh
```diff
@@ -9,25 +9,17 @@ LANGUAGES=( "Base" )
...
 
     # Extract localizable strings from .swift files
     find ${NAVIGATION} -name "*.swift" -print0 | xargs -0 xcrun extractLocStrings -o "${NAVIGATION}/Resources/${lang}.lproj"
     STRINGS_FILE="${NAVIGATION}/Resources/${lang}.lproj/Localizable.strings"
+    convertIfNeeded "$STRINGS_FILE"
 
     # Extract localizable strings from .swift files
     find ${CORE} -name "*.swift" -print0 | xargs -0 xcrun extractLocStrings -o "${CORE}/Resources/${lang}.lproj"
     STRINGS_FILE="${CORE}/Resources/${lang}.lproj/Localizable.strings"
-
```
michaelkirk, line 21: I removed some strings, so I wanted to regenerate them. Unfortunately the script was broken, so I fixed that too.

It seems like maybe there were some other stale strings that hadn't been removed by earlier work.

MapboxNavigation/UIView.swift
```diff
@@ -70,7 +70,10 @@ extension UIView {
     }
     
     func pinInSuperview(respectingMargins margins: Bool = false) {
-        guard let superview else { return }
+        guard let superview else {
+            assertionFailure("superview was unexpectedly nil")
```
michaelkirk, line 74: I added this while debugging. It seems like a good check, but it's ancillary to my change. I can remove it if anyone is offended.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -414,11 +416,6 @@ class RouteMapViewController: UIViewController {
         self.navigationView.endOfRouteView = endOfRoute.view
         self.navigationView.constrainEndOfRoute()
         endOfRoute.didMove(toParent: self)
-
-        endOfRoute.dismissHandler = { [weak self] _, _ in
```
michaelkirk, line 418: Now that we're building programmatically, not from the storyboard, this moved to the initializer.

MapboxNavigation/EndOfRouteViewController.swift
```diff
@@ -69,137 +87,58 @@ class EndOfRouteViewController: UIViewController {
...
+    override func loadView() {
+        self.view = EndOfRouteContentView.forAutoLayout()
+        view.backgroundColor = .white
+        self.view.addSubview(self.column)
+        self.activateLayoutConstraints()
     }
     
     override func viewWillAppear(_ animated: Bool) {
-        super.viewWillDisappear(animated)
-        view.roundCorners([.topLeft, .topRight])
```
michaelkirk, line 83: After converting to a programmatic view, the view was not visible unless I commented out `view.roundCorners` here.

`view.roundCorners` depends on `view.bounds`, which are not necessarily at their final state in `viewWillAppear`, so I've moved it to viewDidLayoutSubiews, which guarantees the views bounds are set. 

Also this means we'll correctly re-compute the path if the bounds change (e.g. on device rotation).

MapboxNavigation/EndOfRouteViewController.swift
```diff
@@ -69,137 +87,58 @@ class EndOfRouteViewController: UIViewController {
...
-        guard let name = destination?.name?.nonEmptyString else { return self.styleForUnnamedDestination() }
-        self.primary.text = name
-    }
-
-    private func clearInterface() {
-        self.primary.text = nil
-        self.stars.rating = 0
-    }
-    
-    private func styleForUnnamedDestination() {
```
michaelkirk, line 168: I inlined this because it was short, and easier to understand the parallel logic of the "else" case.

For example, we now set `self.staticYouHaveArrived.alpha` back to `1.0` in the case that a new destination has a name after a previous one doesn't.

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -882,70 +874,6 @@ extension RouteMapViewController: StepsViewControllerDelegate {
...
-        case .easeIn:
-            self = .curveEaseIn
-        case .easeOut:
-            self = .curveEaseOut
-        case .easeInOut:
-            self = .curveEaseInOut
-        case .linear:
-            self = .curveLinear
-        @unknown default:
-            fatalError("Unknown curve")
```
michaelkirk, line 944: There was a crash here, fixed in 2bf7799f, but then since it was only used for the feedback views, I deleted it all anyway.

MapboxNavigation/EndOfRouteViewController.swift
```diff
@@ -69,137 +87,58 @@ class EndOfRouteViewController: UIViewController {
...
-        self.commentViewContainer.alpha = 0.0 // setting initial hidden state
+    override func loadView() {
+        self.view = EndOfRouteContentView.forAutoLayout()
+        view.backgroundColor = .white
+        self.view.addSubview(self.column)
+        self.activateLayoutConstraints()
     }
     
     override func viewWillAppear(_ animated: Bool) {
-        super.viewWillDisappear(animated)
```
michaelkirk, line 82: I think this was a typo?: **Dis**appear

MapboxCoreNavigation/RouteController.swift
```diff
@@ -424,19 +424,23 @@ extension RouteController: CLLocationManagerDelegate {
...
 
-        if self.routeProgress.currentLegProgress.remainingSteps.count <= 1, remainingVoiceInstructions.count == 0, currentDestination != self.previousArrivalWaypoint {
-            self.previousArrivalWaypoint = currentDestination
+        guard self.routeProgress.currentLegProgress.remainingSteps.count <= 1 else { return }
 
-            self.routeProgress.currentLegProgress.userHasArrivedAtWaypoint = true
+        if let remainingVoiceInstructions = routeProgress.currentLegProgress.currentStepProgress.remainingSpokenInstructions {
```
boldtrn, line 430: If I understand this right, if we have no voiceInstructions, then we will progress the legIndex on the second to last step, which might be very long.

To be honest I don't like the previous code here as well.

How about we rewrite this code to:

- Check if user is on the second to last step
- If the delegate is implemented, we ask the delegate: self.delegate?.routeController?(self, didArriveAt: currentDestination)
- If the delegate is not implemented, I would propose something like (distance remaining to next WP < 50m or duration remaining < 5s or something like this and then advance to next leg. 

Unless I am missing anything, we should also only set 

```
        self.previousArrivalWaypoint = currentDestination
        self.routeProgress.currentLegProgress.userHasArrivedAtWaypoint = true
```

If the delegate returns true.

boldtrn:
Thanks a lot for looking into this @michaelkirk. I would really appreciate if you could split this into two PRs. 

IMHO, the RouteController change is very critical, so I'd appreciate that in a separate PR, ideally with some tests that make sure everything is working as expected.

Since we are not using the RouteMapViewController in our app, I am 100% sure about these changes and maybe @hactar might @Patrick-Kladek would also like to review these.

Patrick-Kladek:
Regarding this, it would probably be a good idea to release this as 3.1 or 3.0.1 as this is needed for the old navigation. Once we release #54 this is no longer needed and would be the responsibility of the app to present such a screen.

boldtrn:
Yes makes sense, so you are fine if we don't merge #54 today, first get this fix through, release a new version and thne merge #54?

Patrick-Kladek:
@boldtrn yes, I don't mind if something takes a day longer to make it right. So lets try to get this done first.

Patrick-Kladek:
@michaelkirk any update on this?

michaelkirk:
> IMHO, the RouteController change is very critical, so I'd appreciate that in a separate PR, ideally with some tests that make sure everything is working as expected.

The RouteController changes by themselves make sure that the EORVC is presented.

But separate bugs (also fixed in this PR) will cause your app to crash when that EORVC is presented. 

The result of fixing the first part without fixing the second part means more apps will crash than without any changes — that's why I coupled them in this big PR.

Am I correct that you want me to make a PR that will fix only the presentation logic, resulting in more apps crashing until further changes to the EORVC are also merged?

MapboxCoreNavigation/RouteController.swift
michaelkirk: > If I understand this right, if we have no voiceInstructions, then we will progress the legIndex on the second to last step, which might be very long.

Yes, that's the pre-existing logic. If you're saying it's weird, I agree with you. Maybe it was never relevant before because mapbox always included voice instructions? I honestly don't know.

> If the delegate is implemented, we ask the delegate: self.delegate?.routeController?(self, didArriveAt: currentDestination)

Just in case there is confusion here, on iOS, the `did` in a method name connotes a hook called after something has already happened.

e.g. `viewDidAppear` is a hook called after the view appears, `application(_:didReceiveRemoteNotification:)` is a hook called after the application receives a remote notification. Contrast it with `will` which is called just before the thing happens. 

- `viewWillAppear`  <- user hook
-  (*view appears*)  <- system does this
- `viewDidAppear` <- user hook

re: `didArriveAt:`

>  - returns: True to advance to the next leg, if any, or false to remain on the completed leg.

So the existing boolean returned by `didArriveAt ` does not answer the question "did I arrive?". It says "you have arrived. Should I continue navigating you to the next leg?". 

I guess it's only relevant to multi-leg journeys. TBH I haven't used it.

I agree that the current logic for determining when you're considered to have have arrived is weird. I think your approach, to compare remaining distance/time, makes much more sense than what's there now. I will update that logic. 

I don't think it's worth exposing a user over-ridable delegate method to customize this behavior. Though, I don't feel strongly about that if you do. I *do* feel strongly that **iff** we create and expose such a delegate method, it should be called something like `shouldArriveAt:` (as opposed to `did...`)

boldtrn: > Maybe it was never relevant before because mapbox always included voice instructions?

Yes, I think so.

> So the existing boolean returned by didArriveAt does not answer the question "did I arrive?". It says "you have arrived. Should I continue navigating you to the next leg?".

Ah ok got it, yeah I still trown off by iOS conventions, thanks.

> I don't think it's worth exposing a user over-ridable delegate method to customize this behavior

Ok, maybe we just make this whole function overwritable? As, to be honest we are thinking about overwriting this with some more specialized code in our app ;)

boldtrn:
> Am I correct that you want me to make a PR that will fix only the presentation logic, resulting in more apps crashing until further changes to the EORVC are also merged?

My idea was to fix the bug in the EORVC first, as this bug seems to be concerning all users that use it?

The change in the RouteController could be independent, but doesn't have to be, I was just a bit concerned that I might block this too long as I really care about these few lines, as they are a crucial core of the navigation experience and need to work well :)

MapboxCoreNavigation/RouteController.swift
michaelkirk: I've pushed up a commit that implements this. It turns out there was already some "slop" built in. See https://github.com/maplibre/maplibre-navigation-ios/pull/58/commits/a16516771754e76b6b27e8ae688f8d9c9d3bd7d2

>> I don't think it's worth exposing a user over-ridable delegate method to customize this behavior
>
> Ok, maybe we just make this whole function overwritable? As, to be honest we are thinking about overwriting this with some more specialized code in our app ;)

Heh, I didn't think it was worthwhile because I was having a hard time imagining that someone would have a substantially different definition of what "arrival" means, but you have already proven me wrong! 😆 I'm happy to add a delegate method in that case.

Can you share what your definition of arrival would look like? I'm curious.

MapboxCoreNavigation/RouteController.swift
```diff
@@ -425,7 +425,9 @@ extension RouteController: CLLocationManagerDelegate {
     func updateRouteLegProgress(for location: CLLocation) {
         let currentDestination = self.routeProgress.currentLeg.destination
 
-        guard self.routeProgress.currentLegProgress.remainingSteps.count <= 1 else { return }
+        guard self.routeProgress.currentLegProgress.remainingSteps.count == 0 else {
```
boldtrn, line 428: We should not do this :), can we use distance or time remaining? For this to become true, the arrival needs to be exactly triggered, which is especially an issue for multi leg routes.

boldtrn: Ah, sorry, I just saw your comment. To be honest, I would prefer if we could actually create another variable like the RouteControllerManeuverZoneRadius, but for leg progress distance.

MapboxCoreNavigation/RouteController.swift
boldtrn: We mostly have multi leg routes, so we might want to do something different for the last waypoint compared to the others, or something like that :)

MapboxCoreNavigation/RouteController.swift
michaelkirk: `RouteControllerManeuverZoneRadius` is used 5 times in the code base, and it seems to be used pretty consistently to say:

> How close do we need to be to a thing to be considered "at" that thing?

TBH, that seems exactly in line with how we're using it here. Why would we want it to be different ? Do you actually have this use case? What value would you choose for it?

Examples pulled from the code:
---

Are we "at" the beginning of a leg:
```
let isWithinDepartureStep = distanceToFirstCoordinateOnLeg < RouteControllerManeuverZoneRadius
```

Are we "at" the intersection:
```
if absoluteDistanceToIntersection <= RouteControllerManeuverZoneRadius {
```

Are we at/along the current step?
```
// If user is close to an intersection, the radius will be lower, so reroutes will fire easier.
let radius = max(reroutingTolerance, RouteControllerManeuverZoneRadius)
```

Are we "at" the end of a step?
```
if userSnapToStepDistanceFromManeuver <= RouteControllerManeuverZoneRadius,
```

Are we *not* "at" the start of the route?
```
lastKnownUserAbsoluteDistance > RouteControllerManeuverZoneRadius
```

michaelkirk:
I'm going to close this. #54 was a really wide change, so a simple rebase is not feasible.

PR #59, opened 2024-06-13
Make sure the puck stays on the route while animating the map

boldtrn:
This is a followup for #27. While the puck stays on the route line while manually panning the map, it still veers of when running animations (like zoom to the whole route). For these kinds of animations, implementers can now call the `enableFrameByFrameCourseViewTracking` from outside of the `NavigationMapView` to fix special cases like this.

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -317,14 +317,17 @@ open class NavigationMapView: MLNMapView, UIGestureRecognizerDelegate {
         }
     }
     
-    // Track position on a frame by frame basis. Used for first location update and when resuming tracking mode
-    func enableFrameByFrameCourseViewTracking(for duration: TimeInterval) {
+    // MARK: - User Tracking
+    
+    /**
+     Track position on a frame by frame basis. Used for first location update and when resuming tracking mode.
+     Call this method when you are doing custom zoom animations, this will make sure the puck stays on the route during these animations.
+     */
+    @objc public func enableFrameByFrameCourseViewTracking(for duration: TimeInterval) {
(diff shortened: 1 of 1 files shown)


michaelkirk:
To be sure I understand: This change by itself won't fix anything, but by making the method public, others can call it when they do things that would otherwise break the animation.

Is that correct? That seems reasonable if so.

But I'm a little confused by your "zoom to the whole route" example.
I'm assuming you mean the codepath that calls `RouteMapViewController.toggleOverview(_)` - right? 

It actually already calls this (previously internal) internal method:

```swift
@objc func toggleOverview(_ sender: Any) {
    self.mapView.enableFrameByFrameCourseViewTracking(for: 3) // <-- ALREADY CALLED
    if let coordinates = routeController.routeProgress.route.coordinates, let userLocation = routeController.locationManager.location?.coordinate {
        self.mapView.setOverheadCameraView(from: userLocation, along: coordinates, for: self.overheadInsets)
    }
    self.isInOverviewMode = true
}
```

I haven't really been able to reproduce the puck going off the rails with the overview button, but I did just now file #61, which I think is a separate issue.

boldtrn:
> To be sure I understand: This change by itself won't fix anything, but by making the method public, others can call it when they do things that would otherwise break the animation.

Yes correct.

> But I'm a little confused by your "zoom to the whole route" example.

Yeah, it's just an arbitrary example. In our app we zoom out to show a set of multiple points and the user location. Essentially any camera animation would need something like this. And if you are using something that is not yet included in the NavigationMapView (or RouteMapViewController), you can't use the method yet, so I think it makes sense to make it available.

michaelkirk (approved):
>  In our app we zoom out to show a set of multiple points and the user location.

Ah ok, so you're not using the built in "show overview" codepath. That explains my confusion.

LGTM

PR #64, opened 2024-06-14
Only require background audio when using speech synthesis

### LINKED ISSUE #63: Make background Audio (and RouteVoiceController?) optional
## Motivation

I'm not currently using voice instructions. I'd like to add it at some point, but my app is still useful without this feature.

The SDK requires the background "Audio" entitlement at runtime, but without the voice instructions, there's no reason for the background "Audio" entitlement. 

The AppStore may not accept an application with the background audio entitlement that doesn't actually use background audio. (speaking from experience here 😬)

RouteVoiceController.verifyBackgroundAudio is called upon init, which happens when instantiating a NavigationViewController.

## Design Alternatives

1. Make the RouteVoiceController optional so that it's not initialized at all. This will have broaded implications and likely conflict with #54 

2. Keep the RouteVoiceController in place, but only call verifyBackgroundAudio once we actually try to use audio — during speech synthesis.

3. Keep the RouteVoiceController in place, but never call `verifyBackgroundAudio`, or reduce the assert to a logging statement. This runs the risk of the user's app not working the way they expect.

## Design

My preference is alternative 2. It's a small change.

michaelkirk:
Fixes #63

I'm not currently using voice instructions. I'd like to add it at some point, but my app is still useful without this feature.

The SDK requires the background "Audio" entitlement at runtime, but without the voice instructions, there's no reason for the background "Audio" entitlement.

The AppStore may not accept an application with the background audio entitlement that doesn't actually use background audio. (speaking from experience here 😬)

RouteVoiceController.verifyBackgroundAudio is called upon init, which happens when instantiating a NavigationViewController. Instead we should wait until we actually use the audio before verifying this permission.

### DIFF
--- a/MapboxNavigation/RouteVoiceController.swift
+++ b/MapboxNavigation/RouteVoiceController.swift
@@ -86,8 +86,6 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
     override public init() {
         super.init()
 
-        self.verifyBackgroundAudio()
-
         self.speechSynth.delegate = self
         
         self.resumeNotifications()
@@ -193,6 +191,9 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
      - parameter ignoreProgress: A `Bool` that indicates if the routeProgress is added to the instruction.
      */
     open func speak(_ instruction: SpokenInstruction, with locale: Locale?, ignoreProgress: Bool = false) {
+        // Don't require background audio unless the implementer is actually using speech APIs
+        self.verifyBackgroundAudio()
+
         if self.speechSynth.isSpeaking, let lastSpokenInstruction {
             self.voiceControllerDelegate?.voiceController?(self, didInterrupt: lastSpokenInstruction, with: instruction)
         }


MapboxNavigation/RouteVoiceController.swift
```diff
@@ -193,6 +191,9 @@ open class RouteVoiceController: NSObject, AVSpeechSynthesizerDelegate {
      - parameter ignoreProgress: A `Bool` that indicates if the routeProgress is added to the instruction.
      */
     open func speak(_ instruction: SpokenInstruction, with locale: Locale?, ignoreProgress: Bool = false) {
+        // Don't require background audio unless the implementer is actually using speech APIs
+        self.verifyBackgroundAudio()
```
michaelkirk, line 195: Note that this ultimately calls an `assertionFailure` so it shouldn't have any bearing on release builds one way or the other.

boldtrn (approved):
LGTM, thanks 🎉

michaelkirk:
Rebased.

This has been approved for a few days and is pretty small, so I'm going to merge. LMK if there's a different process I should follow in the future.

PR #65, opened 2024-06-14
Clean up StyleManager and respect dynamic type

michaelkirk:
### Description

I broke this into two commits, it might be worth reviewing the commits separately. If you'd prefer, I can open them as two separate PRs.

--- 
The first commit changes the API of the StyleManger, and consequently cleans up its internals, but it's intended to be behaviorally a no-op.

Before:
```
let styleManger = StyleManager()
styleManger.styles = [dayStyle] 
// or
styleManger.styles = [dayStyle, nightStyle] 
```

After:
```
let styleManger = StyleManager(dayStyle: dayStyle)
// or
let styleManger = StyleManager(dayStyle: dayStyle, nightStyle: nightStyle)

styleManager.ensureAppropriateStyle()
```

I'm unsure if it's a good idea to require the explicit `styleManager.ensureAppropriateStyle()` after initialization. Previously this logic happend in `styles.didSet`. 

I could move it into the `StyleManger.init`, but editing global state feels like a weird thing to do in an initializer.

--- 

The second commit fixes a problem: changes to dynamic type were not being reflected until the style changed for some other reason. It looks like this was broken at some point (maybe in e134895db65df4501c8773350c5dbdcd163b9da4)

## Before: font size does not update

https://github.com/maplibre/maplibre-navigation-ios/assets/217057/1da3e484-8852-4c88-a948-6aad8f6fad21

## After: font size updates immediately

https://github.com/maplibre/maplibre-navigation-ios/assets/217057/98ec60b8-0344-4eaa-8d87-f3d1267048a3

## with small font

Screenshot 2024-06-14 at 13 36 47

## with big font

Screenshot 2024-06-14 at 13 36 53

--

Note: I'm currently unable to test on CarPlay. Apparently you need some special entitlement from Apple even to run your app on the CarPlay simulator (is that right?!).

Even though I don't currently plan to use CarPlay in my own app, I've applied for the entitlement so I can test changes to like this that might brush up against it.

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -391,9 +391,8 @@ open class NavigationViewController: UIViewController {
         self.view.addSubview(mapSubview)
         mapSubview.pinInSuperview()
         
-        self.styleManager = StyleManager(self)
-        self.styleManager.styles = [dayStyle, nightStyle]
-        
+        self.styleManager = StyleManager(self, dayStyle: dayStyle, nightStyle: nightStyle)
+
         self.mapViewController.navigationView.hideUI(animated: false)
         self.mapView.tracksUserCourse = false
     }
@@ -419,7 +418,12 @@ open class NavigationViewController: UIViewController {
         self.resumeNotifications()
         self.view.clipsToBounds = true
     }
-    
+
+    override open func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        self.styleManager.ensureAppropriateStyle()
+    }
+
     // MARK: - NavigationViewController
 	
     public func startNavigation(with route: Route, routeController: RouteController? = nil, locationManager: NavigationLocationManager? = nil) {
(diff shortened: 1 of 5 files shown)


boldtrn:
> Note: I'm currently unable to test on CarPlay. Apparently you need some special entitlement from Apple even to run your app on the CarPlay simulator (is that right?!).

I saw you slack message. It seems like you were able to get CarPlay to work, were you able to check this PR?

michaelkirk:
I did get the car play entitlement, but I haven’t had time yet to figure out how to launch the car play view controllers. I’ll figure that out next week. That’d be something good to add to the example app once that’s merged.On Jun 15, 2024, at 01:11, Robin ***@***.***> wrote:﻿

Note: I'm currently unable to test on CarPlay. Apparently you need some special entitlement from Apple even to run your app on the CarPlay simulator (is that right?!).

I saw you slack message. It seems like you were able to get CarPlay to work, were you able to check this PR?

MapboxNavigation/NavigationViewController.swift
```diff
@@ -651,8 +655,13 @@ extension NavigationViewController: RouteControllerDelegate {
...
 extension NavigationViewController: TunnelIntersectionManagerDelegate {
     public func tunnelIntersectionManager(_ manager: TunnelIntersectionManager, willEnableAnimationAt location: CLLocation) {
-        self.routeController?.tunnelIntersectionManager(manager, willEnableAnimationAt: location)
-        self.styleManager.applyStyle(type: .night)
+        guard let routeController else {
+            return
+        }
+        routeController.tunnelIntersectionManager(manager, willEnableAnimationAt: location)
+        // If we're in a tunnel at sunrise, don't let the timeOfDay timer clobber night mode
+        self.styleManager.cancelTimeOfDayTimer()
```
hactar, line 663: Good catch to cancel this when one is in a tunnel. But don't we need to re-add the timer once the user leaves the tunnel?

michaelkirk: Fair question! I should have called this out at the top level, but this is actually not a change in behavior. Previously this logic lived in `applyStyle`. 

I was working on applyStyle and could not figure out why we'd cancel the the timer. Eventually I realized it was only for this case, so I extracted the functionality here.

After you exit the tunnel, same as before, `self.styleManager.timeOfDayChanged()` is called which restarts the timer.


e.g. the green area roughly corresponds to a tunnel over the highway:

https://github.com/user-attachments/assets/236d7081-1462-477f-a3e3-0c02e77f433e

michaelkirk: A little more context in: https://github.com/maplibre/maplibre-navigation-ios/pull/65/commits/f8eb874453750c16a639ba3c0614df9fd19abd6d

PR #71, opened 2024-06-24
Fix crash in EndOfRouteViewController and restore its presentation

michaelkirk:
This is a subset of #58, extracted and rebased now that #54 is merged.

Before #54, upon completion of a route, the EndOfRouteViewController (EORVC) would be presented. It was intentionally removed in #54, but I'd like to revisit that decision. Frankly there was so much going on in #54, that I didn't realize the implications of this particular change. 

I think it's a good idea to include a reasonable "you've finished the route" UI by default. Same as before #54, it can be disabled by setting `vc.showsEndOfRouteFeedback = false`.

Note that there is still a separate bug where the route is only considered "completed" when the route contains audio instructions. As [suggested](https://github.com/maplibre/maplibre-navigation-ios/pull/58#issuecomment-2176941128), I'll follow up with a fix for that separately.

https://github.com/maplibre/maplibre-navigation-ios/assets/217057/d1a505ec-7a8c-4932-9231-d48e859eaba5

For my own App, I've overridden the appearance of the button to be a "big red button", but I left the default one styled basic (as it was before my changes):

Screenshot 2024-06-24 at 13 43 23

If people think this is a better default, I can upstream those style changes too.


MapboxNavigation/NavigationView.swift
```diff
@@ -123,7 +129,19 @@ open class NavigationView: UIView {
...
-    
+
+    var endOfRouteView: UIView? {
+        didSet {
+            if let active: [NSLayoutConstraint] = constraints(affecting: oldValue) {
+                NSLayoutConstraint.deactivate(active)
+            }
+
+            oldValue?.removeFromSuperview()
+            if let eor = endOfRouteView { addSubview(eor) }
```
hactar, line 140:
```suggestion
            if let localEndOfRouteView = endOfRouteView {
                addSubview(localEndOfRouteView) 
            }
```

hactar (approved):
I'd propose a small code cycle change, but nothing blocking.

MapboxNavigation/NavigationView.swift
michaelkirk: Good call — I ended up addressing it slightly differently. Please take another look.

PR #73, opened 2024-06-25
Restore the ability to init NavVC without animations.

michaelkirk:
Prior to #54, the NavigationViewController could only be used modally.

With #54, you can switch between "map mode" and "navigation mode". 

An animation was introduced in `navigationView.showUI` for the sake of smoothly transitioning from "map mode" to "navigation mode".

However, if you are still using the NavigationViewController modally, this animation might interfere with other things you were doing - like zooming before presentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,19 +2,14 @@
 
 ## Unreleased
 * Start & Stop Navigation in existing Map
-    - Removed: `NavigationViewController(for route: Route, dayStyle: Style, routeController: RouteController? = nil, locationManager: NavigationLocationManager? = nil, voiceController: RouteVoiceController? = nil)` use `NavigationViewController(dayStyleURL: URL, nightStyleURL: URL? = nil,directions: Directions = .shared, voiceController: RouteVoiceController = RouteVoiceController())` followed by `startNavigation(with route: Route)` instead.
-    - To simulate a route, pass a `SimulatedLocationManager` to `startNavigation()` function: 
+    - Removed: `NavigationViewController(for route: Route, dayStyle: Style, routeController: RouteController? = nil, locationManager: NavigationLocationManager? = nil, voiceController: RouteVoiceController? = nil)` use `NavigationViewController(dayStyleURL: URL, nightStyleURL: URL? = nil,directions: Directions = .shared, voiceController: RouteVoiceController = RouteVoiceController())` followed by `startNavigation(with route: Route, animated: Bool)` instead.
+    - To simulate a route, pass a `SimulatedLocationManager` to `startNavigation()` function:
 
     ```swift
     let vc = NavigationViewController(dayStyleURL: AppConfig().tileserverStyleUrl)
-
-    if Env.current.simulateLocationForTesting {
-        let simulatedLocationManager = SimulatedLocationManager(route: route)
-        simulatedLocationManager.speedMultiplier = 5
-        vc.startNavigation(with: route, locationManager: simulatedLocationManager)
-    } else {
-        vc.startNavigation(with: route)
-    }
+    let simulatedLocationManager = SimulatedLocationManager(route: route)
+    simulatedLocationManager.speedMultiplier = 5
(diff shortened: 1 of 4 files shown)


MapboxNavigation/NavigationViewController.swift
```diff
@@ -496,8 +496,8 @@ open class NavigationViewController: UIViewController {
                 let directions = routeController.directions
                 let route = routeController.routeProgress.route
                 let navigationViewController = NavigationViewController(dayStyle: DayStyle(demoStyle: ()), nightStyle: NightStyle(demoStyle: ()), directions: directions)
-                navigationViewController.startNavigation(with: route, routeController: routeController, locationManager: locationManager)
-                
+                navigationViewController.startNavigation(with: route, animated: true, routeController: routeController, locationManager: locationManager)
```
hactar, line 499: Why do we pass animated: true in here instead of animated: animated?

CHANGELOG.md
```diff
@@ -2,19 +2,14 @@
...
-    if Env.current.simulateLocationForTesting {
-        let simulatedLocationManager = SimulatedLocationManager(route: route)
-        simulatedLocationManager.speedMultiplier = 5
-        vc.startNavigation(with: route, locationManager: simulatedLocationManager)
-    } else {
-        vc.startNavigation(with: route)
-    }
+    let simulatedLocationManager = SimulatedLocationManager(route: route)
+    simulatedLocationManager.speedMultiplier = 5
+    vc.startNavigation(with: route, animated: false, locationManager: simulatedLocationManager)
```
Patrick-Kladek, line 12: Why did you remove this? This way navigation will always simulate the route, which is helpful while working in the simulator but the opposite of what users want. It's not intuitive to omit this parameter (when the changelog says to pass it) to get default behaviour back. Maybe a comment here would be better suited

MapboxNavigation/NavigationViewController.swift
```diff
@@ -422,7 +422,7 @@ open class NavigationViewController: UIViewController {
     
     // MARK: - NavigationViewController
 	
-    public func startNavigation(with route: Route, routeController: RouteController? = nil, locationManager: NavigationLocationManager? = nil) {
+    public func startNavigation(with route: Route, animated: Bool, routeController: RouteController? = nil, locationManager: NavigationLocationManager? = nil) {
```
Patrick-Kladek, line 425:
```suggestion
    public func startNavigation(with route: Route, animated: Bool = true, routeController: RouteController? = nil, locationManager: NavigationLocationManager? = nil) {
```

MapboxNavigation/NavigationViewController.swift
Patrick-Kladek: This is called in 

```swift
func carPlayManager(_ carPlayManager: CarPlayManager, didBeginNavigationWith routeController: RouteController, window: UIWindow) 
```

It's hidden in github if you don't show more. I think passing animated: true is the right choice here.

Patrick-Kladek (approved):
The code looks good. I will approve it immediately; however, please address my comments.

CHANGELOG.md
michaelkirk: I thought this was only supposed to be an example for the immediately preceding bullet point, which was about simulating location. My mistake! I'll revert it.

MapboxNavigation/NavigationViewController.swift
michaelkirk: I typically use default values only when people will almost always want that default value. 

To me this feels closer to 50/50, so I'd prefer to have people type it out rather than run the risk of people assuming it doesn't exist. For example, anyone using the navigation VC modally (e.g. any legacy users), will want this to be false.

Patrick-Kladek: Okay makes sense, then let's keep it without the default parameter 👍

Patrick-Kladek (approved):

hactar (approved):

PR #76, opened 2024-07-01
SwiftUI Optimisation

Patrick-Kladek:
### Description

We are wrapping MapLibre-navigation in a `UIViewControllerRepresentable`. The `updateUIViewController` function is run multiple times leading to setting the same variable over and over.

### Open Tasks

- [x] prevent code from running again after setting the same property again
- [x] only call `navigationViewControllerDidFinishRouting` when confirmed via button

### DIFF
--- a/MapboxNavigation/NavigationViewController.swift
+++ b/MapboxNavigation/NavigationViewController.swift
@@ -214,6 +214,8 @@ open class NavigationViewController: UIViewController {
 	
     var currentStatusBarStyle: UIStatusBarStyle = .default {
         didSet {
+            guard oldValue != self.currentStatusBarStyle else { return }
+
             self.mapViewController.instructionsBannerView.backgroundColor = InstructionsBannerView.appearance().backgroundColor
             self.mapViewController.instructionsBannerContentView.backgroundColor = InstructionsBannerContentView.appearance().backgroundColor
         }
@@ -228,6 +230,8 @@ open class NavigationViewController: UIViewController {
      */
     public var route: Route? {
         didSet {
+            guard oldValue != self.route else { return }
+			
             if let route {
                 if self.routeController == nil {
                     let routeController = RouteController(along: route, directions: self.directions, locationManager: self.locationManager)
@@ -283,6 +287,8 @@ open class NavigationViewController: UIViewController {
      */
     public var routeController: RouteController? {
         didSet {
+            guard oldValue != self.routeController else { return }
+			
             self.mapViewController.routeController = self.routeController
         }
     }
@@ -313,6 +319,8 @@ open class NavigationViewController: UIViewController {
      */
     public var automaticallyAdjustsStyleForTimeOfDay = true {
         didSet {
+            guard oldValue != self.automaticallyAdjustsStyleForTimeOfDay else { return }
+			
             self.styleManager.automaticallyAdjustsStyleForTimeOfDay = self.automaticallyAdjustsStyleForTimeOfDay
         }
     }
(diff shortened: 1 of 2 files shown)


MapboxNavigation/NavigationViewController.swift
```diff
@@ -649,7 +658,9 @@ extension NavigationViewController: RouteControllerDelegate {
                 guard let self else {
                     return
                 }
-                self.delegate?.navigationViewControllerDidFinishRouting?(self)
+                self.mapViewController.hideEndOfRoute { _ in
+                    self.delegate?.navigationViewControllerDidFinishRouting?(self)
+                }
```
Patrick-Kladek, line 663: I've ensured that this is called regardless of `showsEndOfRouteFeedback` Is set to true or false.

MapboxNavigation/NavigationViewController.swift
```diff
@@ -467,6 +475,7 @@ open class NavigationViewController: UIViewController {
 
         self.voiceController = nil
         self.route = nil
+        self.locationManager = NavigationLocationManager()
```
Patrick-Kladek, line 478: Resets LocationManager in case a `SimulatedLocationManager` was used

michaelkirk:
Hi @Patrick-Kladek - 
I just want to make sure that it stays focused on one issue, so that we can avoid having a far reaching PR that changes a lot of things at once. It makes it hard to test and evaluate. I see this is currently a draft. Why is that? What's the scope of this PR?

Patrick-Kladek:
@michaelkirk Sure, this PR is about fixing small things we've missed in #54. It has 3 small changes, I hope you don't want me to submit 3 PRs 😅, that's why I included it here. I've already got the approval from my team, so it's ready to review here too.

hactar (dismissed):

michaelkirk:
> It has 3 small changes, I hope you don't want me to submit 3 PRs 😅

Well when you phrase it like that it makes me feel crazy 😆, but to be honest, it certainly would be my preference if they are separate! It makes it easier to talk about separate concerns.

For example, I have no concerns with:

https://github.com/maplibre/maplibre-navigation-ios/pull/76/commits/b0c1397a7e462476b2fb2e22895545493a041601

But https://github.com/maplibre/maplibre-navigation-ios/pull/76/commits/41116711a2e54e108422bd4f8080d723e7a8625f seems like the wrong fix. Since location manger is per-route, not per-view-controller, it probably makes more sense to set it in `startRoute` rather than in `init`, right?

And as for https://github.com/maplibre/maplibre-navigation-ios/pull/76/commits/427721ac9163c556eee7260598a0dc46a7019df6, In my own application, I was calling hideEndOfRoute in my application logic, but I think you're right that we should probably include this in the library. However, I don't think it makes sense to call `hideEndOfRoute` unless `showEndOfRoute` was called, so probably here is the right place to include it?
https://github.com/maplibre/maplibre-navigation-ios/blob/fa4d5f46b614fe39000cd0d02276fc92099e85ed/MapboxNavigation/RouteMapViewController.swift#L450

What do you think?

Patrick-Kladek:
Good point that the LocationManager is per Route, I fixed that.

`hideEndOfRoute` has no side effects, it will simply remove the built-in feedback view if it's in the view hierarchy. If you want to provide your own, you do it differently anyway by listening to the delegate method `navigationViewControllerDidFinishRouting()`.

I checked that the `navigationViewControllerDidFinishRouting()` is called regardless of the `showsEndOfRouteFeedback` property.

michaelkirk (approved):
> Good point that the LocationManager is per Route, I fixed that.

👍 

> hideEndOfRoute has no side effects, it will simply remove the built-in feedback view if it's in the view hierarchy. If you want to provide your own, you do it differently anyway by listening to the delegate method navigationViewControllerDidFinishRouting().

Ok, I don't feel that strongly about this one.

PR #81, opened 2024-07-11
Fix version 3 documentation

hactar:
We decided to do a last minute release before releasing a version with "Start navigation in existing map" but forgot to update the documentation, so the documentation makes it sound like this feature is in version 3, which it is not. This PR updates the documentation.


Patrick-Kladek (approved):

michaelkirk (approved):

michaelkirk:
Thanks!

Issue #80, opened 2024-07-10
Navigation Puck is centered, but should be in bottom third so you can see farther ahead.


michaelkirk:
When in "navigation" mode, the camera perspective shifts to be slightly "behind" the user so you can look ahead towards the direction you're heading. 

### Steps to Trigger Behavior

1. In the example app (🎉) start navigating
2. See that the puck is in the middle of the screen

### Expected Behavior

The puck should be in the bottom 1/3 of the view port so you can see farther ahead.

### Actual Behavior

The puck is in the center of the screen.

### Screenshots (if applicable)

<img width=300 src=https://github.com/maplibre/maplibre-navigation-ios/assets/217057/913abeca-a074-4c12-bd35-1b918a9a272f>

### Version(s) affected

- maplibre-navigation-ios Version HEAD: https://github.com/maplibre/maplibre-navigation-ios/commit/db8dd912771a6b5a6dae0f55f98667c70afc9ae3

PR #82, opened 2024-07-12
Fix puck location during navigation

Patrick-Kladek:
### Description

~**Warning**: do not merge right away, this contains changes from #76 which needs to be merged first!~

For some reason, the user puck was centered on the map. This reduced the visible road ahead and made navigation quite bad.

### Open Tasks

- [x] Fix user puck position during navigation

### Infos for Reviewer

The trick was to create a padding close to the region where the user puck is positioned. This padding is then used when the camera is set for the MapView.

Here is a recording from our app (SwiftUI), while starting the navigation and transitioning between overhead mode. It's not perfect yet (the overhead move is not fully visible, need to investigate further) but it's a lot better than what we currently have. There is no workaround needed to get the puck down and it is also butter smooth.

https://github.com/user-attachments/assets/52370de4-a20d-4246-8a46-5115f216ca32


michaelkirk:
Is this about https://github.com/maplibre/maplibre-navigation-ios/issues/80 or something else?

Patrick-Kladek:
Yes, I wasn't aware that you created an issue already

hactar:
@michaelkirk could you take a look at this? You recently did some "puck work" so I'd assume you're better suited then me. :)

Example/example/SceneDelegate.swift
```diff
@@ -101,12 +101,24 @@ private extension SceneDelegate {
...
         guard let waypoint = self.waypoints.randomElement() else { return }
-
+        
         let distance = CLLocationDistance.random(in: 10 ... 100_000)
-        self.viewController.mapView.camera = .init(lookingAtCenter: waypoint.coordinate,
-                                                   acrossDistance: distance,
-                                                   pitch: 0,
-                                                   heading: 0)
+        
+        let userAnchorPoint = self.viewController.mapView.userCourseView!.center
```
michaelkirk, line 107: What is the purpose of this change to the example app? Is it required that every implementor will need to have code like this to have the puck positioned in the bottom 1/3 of the map?

MapboxNavigation/CarPlayNavigationViewController.swift
```diff
@@ -169,7 +169,7 @@ public class CarPlayNavigationViewController: UIViewController, MLNMapViewDelega
                     return
                 }
                 self.mapView?.enableFrameByFrameCourseViewTracking(for: 3)
-                self.mapView?.setOverheadCameraView(from: userLocation, along: self.routeController.routeProgress.route.coordinates!, for: self.edgePadding)
+                self.mapView?.setOverheadCameraView(from: userLocation, along: self.routeController.routeProgress.route.coordinates!, insets: self.edgePadding)
```
michaelkirk, line 172: :+1: better name!

MapboxNavigation/NavigationMapView.swift
```diff
@@ -347,20 +355,18 @@ open class NavigationMapView: MLNMapView, UIGestureRecognizerDelegate {
             return
         }
         
-        if !self.tracksUserCourse || self.userAnchorPoint != userCourseView?.center ?? self.userAnchorPoint {
```
michaelkirk, line 350: Why did this boolean check go away?  (TBH I'm not sure I understand what it was doing in the first place)

MapboxNavigation/RouteMapViewController.swift
```diff
@@ -567,6 +568,7 @@ extension RouteMapViewController: NavigationViewDelegate {
         if userTrackingMode == .none, !self.isInOverviewMode {
             self.navigationView.wayNameView.isHidden = true
         }
+        mapView.userTrackingMode = userTrackingMode
```
michaelkirk, line 571: Why did this change?

michaelkirk (commented):
Sorry for the slow review! I'm not very familiar with the puck placement code, so I mostly just have questions at this point.

Example/example/SceneDelegate.swift
Patrick-Kladek: No this has nothing to do with moving the puck down therefore it's not required by every implementor. 

This is a function in the sample project that randomly positions the camera for testing purposes. I wanted to test what happens when someone interferes with the camera during navigation - more specifically, what can go wrong if Maplibre-DSL (a SwiftUI wrapper) decides to reposition the camera. This allows to test this functionality while not embedded in SwiftUI.

MapboxNavigation/NavigationMapView.swift
Patrick-Kladek: I think this check was in place so centering the puck only happened if it was not centred already (optimisation maybe). As the map is moving during navigation we need to position the puck anyway, thats why I removed it.

MapboxNavigation/RouteMapViewController.swift
Patrick-Kladek: Because otherwise I would need to set the userTracking mode in `startNavigation()`. This way is more reliable. 

If you look at the whole function there seems to be something missing, `userTrackingMode` is read from the mapView, then updated but never written back to the mapView, so it's basically unused. I suspect this was the intent in the first place but could have changed in a previous PR (maybe on accident).

michaelkirk: I'm not understanding why this would need to change to address the "position camera in bottom 1/3" of the screen logic. Is it required or is this some bonus work?

MapboxNavigation/NavigationMapView.swift
michaelkirk: The map stops moving if the user pans. Does that matter? 

Is this change required for the "position camera in bottom 1/3" of the screen work, or is this some bonus cleanup?

Example/example/SceneDelegate.swift
michaelkirk: Ah ok, I never understood what this button was for. I'm still not sure that I do. Is it supposed to simulate a user interaction to drag the map around? Or are you trying to simulate some rogue software triggered camera movement? And did something about your changes to get the puck positioned in the bottom 1/3 of the screen cause this to behave differently?

In any case a comment on the method is worthwhile.

MapboxNavigation/RouteMapViewController.swift
Patrick-Kladek: This is done as a precaution to sync the tracking mode when navigating. Why else would we read, then modify the `trackingMode` but not assign it to the mapView? You could even say it fixes something a previous PR missed, otherwise we could remove this function altogether as it does nothing.

Bildschirmfoto 2024-08-01 um 14 22 26

michaelkirk: I'm not saying you're wrong, but to review your code I need to understand what you are __trying__ to do.

You said this pr is about positioning the puck in the bottom 1/3 of the viewport.

But this change is about something else, right? Did you observe a broken behavior with this? Or is this just about fixing code that "looked wrong"?

MapboxNavigation/NavigationMapView.swift
michaelkirk: But why did it change in this PR? Again, I can't tell if you are doing random cleanup to make things nicer, or if this is required due to other changes you made to get the puck to appear in the bottom 1/3.

Patrick-Kladek: I did remove this because it makes no difference. This if clause resolved to true in all scenarios I've tested. I guess this was relevant when we manually posited the userCourseView.center to the screen's center. Now that this is gone it's not necessary anymore.

MapboxNavigation/RouteMapViewController.swift
Patrick-Kladek: If you are bothered by this line and want me to remove it because you think it's unnecessary then please say that. I've explained to you twice why I think we need this.

michaelkirk:
> If you are bothered by this line and want me to remove it because you think it's unnecessary then please say that. I've explained to you twice why I think we need this.

I’m not bothered - just trying to understand so I can review your code.

hactar (approved):

PR #85, opened 2024-07-17
Open RouteController to custom location snapping

sotomski:
### Description

This PR opens the `RouteController.snappedLocation` to be additionally modified by delegate implementers.
This is useful to plug in a custom snapping logic.

### Infos for Reviewer

The intention was to open up the snapping logic but make it optional and don't change snapping behavior if custom logic is not provided. If custom snapping doesn't provide a valid value (ie. fails), the default snapping shall be used as fallback.

### DIFF
--- a/MapboxCoreNavigation/RouteController.swift
+++ b/MapboxCoreNavigation/RouteController.swift
@@ -230,7 +230,13 @@ open class RouteController: NSObject, Router {
      - important: If the rawLocation is outside of the route snapping tolerances, this value is nil.
      */
     var snappedLocation: CLLocation? {
-        self.rawLocation?.snapped(to: self.routeProgress.currentLegProgress)
+        let snappedDefault = self.rawLocation?.snapped(to: self.routeProgress.currentLegProgress)
+        guard let raw = self.rawLocation else { return snappedDefault }
+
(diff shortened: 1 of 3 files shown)


MapboxCoreNavigation/RouteController.swift
```diff
@@ -230,7 +230,13 @@ open class RouteController: NSObject, Router {
      - important: If the rawLocation is outside of the route snapping tolerances, this value is nil.
      */
     var snappedLocation: CLLocation? {
-        self.rawLocation?.snapped(to: self.routeProgress.currentLegProgress)
+        let snappedDefault = self.rawLocation?.snapped(to: self.routeProgress.currentLegProgress)
```
boldtrn, line 233: Aren't you doing duplicated work here? If there is a delegate, the delegate should do the snapping, otherwise the default snapping happens.

boldtrn (changes requested):
Thanks for the nice PR, I think this will be helpful in using the library. Could you double check that we don't have to snap the location twice?

boldtrn (approved):
LGTM 👍 - could you add a changelog entry for your changes?

sotomski:
Sure thing 👍

PR #89, opened 2024-07-27
Update MapboxDirections

Patrick-Kladek:
### Description

In the last Maplibre Meeting we decided to use & update MapboxDirections to the original Mapbox version, away from this fork https://github.com/flitsmeister/mapbox-directions-swift.

We also have a Github discussion about this here: https://github.com/maplibre/maplibre-navigation-ios/discussions/87

### Open Tasks

- [x] update to original MapboxDirections Package
- [x] begin removing objc annotations
- [ ] fix build error

### Infos for Reviewer

The fork is way behind so integrating MapboxDirections is not an easy task. MapboxDirections is now optimized for Swift. One of the major changes is that `VisualInstructionComponent` is now renamed to `VisualInstruction.Component`, an enum with associated values. This is incompatible with how we present banners & spoken instructions. I've tried to make the code work but I've definitely broken something.

This PR state is far from finished or even reviewable. I've fixed around 200 build errors, just so that another 90 appeared. Given that I don't even know if this works I'm stopping here for now, but keeping this PR if anyone wants to continue off of it.

### DIFF
--- a/Example/Example.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Example/Example.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -3,10 +3,10 @@
     {
       "identity" : "mapbox-directions-swift",
       "kind" : "remoteSourceControl",
-      "location" : "https://github.com/flitsmeister/mapbox-directions-swift",
+      "location" : "https://github.com/mapbox/mapbox-directions-swift",
       "state" : {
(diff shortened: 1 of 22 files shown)


boldtrn:
> This PR state is far from finished or even reviewable. I've fixed around 200 build errors, just so that another 90 appeared. Given that I don't even know if this works I'm stopping here for now, but keeping this PR if anyone wants to continue off of it.

Thanks for looking into this. Yes this was exactly what I expected, things got renamed and this has a direct effect on the navigation code. So how would you want to continue?

PR #90, opened 2024-07-29
remove unused code

michaelkirk:
MapboxCoreNavigation/CoreFeedbackEvent.swift is a vestige of the telemetry stuff which was removed in several commits, including 62e6415d6e805c5b4696fbd439615144096f6f01

PR #91, opened 2024-07-29
update turf

michaelkirk:
This is a draft because it is based on #90, so please review that first.

### Description

Taking a smaller bite out of #89, this PR only updates our turf dependency.

### DIFF
--- a/MapboxCoreNavigation/CLLocation.swift
+++ b/MapboxCoreNavigation/CLLocation.swift
@@ -125,8 +125,8 @@ extension CLLocation {
 
         guard let closest = nearByPolyline.closestCoordinate(to: coordinate) else { return nil }
         
-        let slicedLineBehind = LineString(coordinates.reversed()).sliced(from: closest.coordinate, to: coordinates.reversed().last)
-        let slicedLineInFront = nearByPolyline.sliced(from: closest.coordinate, to: coordinates.last)
+        guard let slicedLineBehind = LineString(coordinates.reversed()).sliced(from: closest.coordinate, to: coordinates.reversed().last) else { return nil }
+        guard let slicedLineInFront = nearByPolyline.sliced(from: closest.coordinate, to: coordinates.last) else { return nil }
         let userDistanceBuffer: CLLocationDistance = max(speed * RouteControllerDeadReckoningTimeInterval / 2, RouteControllerUserLocationSnappingDistance / 2)
         
         guard let pointBehind = slicedLineBehind.coordinateFromStart(distance: userDistanceBuffer) else { return nil }
--- a/MapboxCoreNavigation/CLLocation.swift
+++ b/MapboxCoreNavigation/CLLocation.swift
@@ -61,7 +61,7 @@ extension CLLocation {
      Returns a Boolean value indicating whether the receiver is within a given distance of a route step.
      */
     func isWithin(_ maximumDistance: CLLocationDistance, of routeStep: RouteStep) -> Bool {
-        guard let closestCoordinate = Polyline(routeStep.coordinates!).closestCoordinate(to: coordinate) else {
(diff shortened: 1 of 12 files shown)


MapboxCoreNavigation/CLLocation.swift
```diff
@@ -121,12 +123,12 @@ extension CLLocation {
...
-        let nearByPolyline = Polyline(coordinates)
-        
+        let nearByPolyline = LineString(coordinates)
+
         guard let closest = nearByPolyline.closestCoordinate(to: coordinate) else { return nil }
         
-        let slicedLineBehind = Polyline(coordinates.reversed()).sliced(from: closest.coordinate, to: coordinates.reversed().last)
-        let slicedLineInFront = nearByPolyline.sliced(from: closest.coordinate, to: coordinates.last)
+        guard let slicedLineBehind = LineString(coordinates.reversed()).sliced(from: closest.coordinate, to: coordinates.reversed().last) else { return nil }
+        guard let slicedLineInFront = nearByPolyline.sliced(from: closest.coordinate, to: coordinates.last) else { return nil }
```
hactar, line 131: Why was returning nil here not needed before?

MapboxCoreNavigation/RouteProgress.swift
```diff
@@ -368,17 +368,17 @@ open class RouteLegProgress: NSObject {
 
         for (currentStepIndex, step) in remainingSteps.enumerated() {
             guard let coords = step.coordinates else { continue }
-            guard let closestCoordOnStep = Polyline(coords).closestCoordinate(to: coordinate) else { continue }
+            guard let closestCoordOnStep = LineString(coords).closestCoordinate(to: coordinate) else { continue }
             let foundIndex = currentStepIndex + self.stepIndex
 
-            // First time around, currentClosest will be `nil`.
```
hactar, line 374: Is this comment no longer correct?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -1071,7 +1074,9 @@ open class NavigationMapView: MLNMapView, UIGestureRecognizerDelegate {
      */
     @objc public func setOverheadCameraView(from userLocation: CLLocationCoordinate2D, along coordinates: [CLLocationCoordinate2D], for bounds: UIEdgeInsets) {
         self.isAnimatingToOverheadMode = true
-        let slicedLine = Polyline(coordinates).sliced(from: userLocation).coordinates
+        guard let slicedLine = LineString(coordinates).sliced(from: userLocation)?.coordinates else {
```
hactar, line 1077: In which cases does LineString.sliced return nil in this case?

hactar:
Looking fine, just have a few questions so that I fully understand the PR. Please resolve the merge conflict and mark as ready for review.

MapboxNavigation/NavigationMapView.swift
michaelkirk: Same cause as my previous comment - `sliced` will now return `nil` rather than an empty `LineString` in the case that `coordinates` is empty.

The implications of this one are trickier though — it's possible that this might result in some change in behavior. My wager is that it's OK, since passing empty coordinates seems ill formed in the first place. 

And in that case, our response is a no-op, rather than exploding or something scary.

I've just now added a debug assert to build confidence that we're not inadvertently hitting that code path. 

What do you think?

michaelkirk (commented):
Thanks for the review - please take another look @hactar.

hactar (approved):
Thanks for your comments and improvements. Sounds all reasonable to me. Approving. Will merge a day later in case someone else wishes to take a look too in the mean time.

MaartenZonneveld:
@michaelkirk  We think this change broke navigation arrows. Can you please take a look at #106 ?

michaelkirk:
I am no longer using maplibre-navigation-ios, so I won't prioritize looking into this. Feel free to revert the update to turf if that solves things for you.

Issue #60, opened 2024-06-13
Creating binaries from SDK for wrapping into XCFramework fails


mindthefish:
I am working on a fork of the SDK with some closed-source additions. I need the ability to deploy the SDK as a binary. So far I am using the pre-SPM version which I can compile using `xcodebuild archive`, `xcodebuild -create-xcframework` or Carthage. Since the switch to SPM, this option is no longer available, so I try to build the current version (main branch) via `swift build` or `swift create-xc-framework`. Unfortunately I encounter errors when building the dependencies.

### Steps to Trigger Behavior

1. checkout the main branch of this repo
2. archive the targets specified in the Package.swift using `xcodebuild` or `swift` commands
3. bundle the archives to a XCFramework

### Expected Behavior
Build / archive command should complete without errors

### Actual Behavior
I.e. `swift build` leads to error: ".../maplibre-navigation-ios/MapboxNavigationObjC/MLNMapView+MLNNavigationAdditions.h:1:9: fatal error: 'MapLibre/Mapbox.h' file not found
#import <MapLibre/Mapbox.h>
        ^~~~~~~~~~~~~~~~~~~
1 error generated."

### Version(s) affected

- main branch at (commit https://github.com/maplibre/maplibre-navigation-ios/commit/eaee92b308df0ad15832767e4ac1d71df8116211), but I think this applies to all versions since change to SPM.

Has anyone ever tried to build maplibre-navigation-ios as a binary? I want to deploy it as XCFramework binary as part of a Swift package.

michaelkirk:
I hit a similar error when trying to generate documentation: https://github.com/maplibre/maplibre-navigation-ios/issues/52#issuecomment-2146249996

mindthefish:
Sounds like the same problem.

mr-rick:
I'm seeing the same thing as well

mr-rick:
Can you give me a hint or lead me in the right direction to get this going? I'm happy to help here... I'm just not an expert on the subject matter.

This will help us in the flutter world.

Thanks!

-Rick

brandub:
Same issue here. Also needing this for flutter dev.

brandub:
Do you think this PR will solve this issue as well?? https://github.com/maplibre/maplibre-navigation-ios/pull/95

michaelkirk:
I don't expect it to.

mindthefish:
[...]

PR #95, opened 2024-08-20
Cleanup: remove private header

michaelkirk:
### Description

This header was copied in from MapLibre Native to expose a private
method.

Previously this got out of sync and caused a bug:
https://github.com/maplibre/maplibre-navigation-ios/pull/53

Instead we can use the already public delegate method.

Note: The deleted header was importing Maplibre/Mapbox.h, but we are
using that elsewhere, so I moved that import up a level.

### Open Tasks

Related:
https://github.com/maplibre/maplibre-native/issues/2228
https://github.com/maplibre/maplibre-navigation-ios/pull/53

### Infos for Reviewer

There should be no change in behavior from this — just cleanup.

### DIFF
--- a/MapboxNavigation/NavigationMapView.swift
+++ b/MapboxNavigation/NavigationMapView.swift
@@ -258,9 +258,7 @@ open class NavigationMapView: MLNMapView, UIGestureRecognizerDelegate {
         }
     }
    
-    override open func mapViewDidFinishRenderingFrameFullyRendered(_ fullyRendered: Bool, frameEncodingTime: Double, frameRenderingTime: Double) {
-        super.mapViewDidFinishRenderingFrameFullyRendered(fullyRendered, frameEncodingTime: frameEncodingTime, frameRenderingTime: frameRenderingTime)
-        
+    func updateCourseTrackingAfterDidFinishRenderingFrame() {
         guard self.shouldPositionCourseViewFrameByFrame else { return }
         guard let location = userLocationForCourseTracking else { return }
         
--- a/MapboxNavigationObjc/include/MapboxNavigation.h
+++ b/MapboxNavigationObjc/include/MapboxNavigation.h
@@ -7,4 +7,4 @@ FOUNDATION_EXPORT double MapboxNavigationVersionNumber;
 FOUNDATION_EXPORT const unsigned char MapboxNavigationVersionString[];
 
 #import "../MBRouteVoiceController.h"
-#import "../MLNMapView+MLNNavigationAdditions.h"
+#import <MapLibre/Mapbox.h>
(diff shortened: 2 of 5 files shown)


boldtrn (approved):
Thanks for looking into this.

If you tested this and everything still works, I am fine with this change. Maybe we give for example @Patrick-Kladek and @hactar a chance to check as well, but other than that, I am fine with merging 👍

brandub:
I'm hoping this PR solves https://github.com/maplibre/maplibre-navigation-ios/issues/60 as well any thoughts?

MapboxNavigationObjc/include/MapboxNavigation.h
```diff
@@ -7,4 +7,4 @@ FOUNDATION_EXPORT double MapboxNavigationVersionNumber;
 FOUNDATION_EXPORT const unsigned char MapboxNavigationVersionString[];
 
 #import "../MBRouteVoiceController.h"
-#import "../MLNMapView+MLNNavigationAdditions.h"
+#import <MapLibre/Mapbox.h>
```
Patrick-Kladek, line 10: While we are at it, can we rename this file to `MapLibre.h` or is this in another package?

MapboxNavigation/NavigationMapView.swift
```diff
@@ -258,9 +258,7 @@ open class NavigationMapView: MLNMapView, UIGestureRecognizerDelegate {
         }
     }
    
-    override open func mapViewDidFinishRenderingFrameFullyRendered(_ fullyRendered: Bool, frameEncodingTime: Double, frameRenderingTime: Double) {
-        super.mapViewDidFinishRenderingFrameFullyRendered(fullyRendered, frameEncodingTime: frameEncodingTime, frameRenderingTime: frameRenderingTime)
-        
+    func updateCourseTrackingAfterDidFinishRenderingFrame() {
```
Patrick-Kladek, line 261: What's the reason for removing the override and then calling this from elsewhere? For me the override function is more logical

MapboxNavigationObjc/include/MapboxNavigation.h
michaelkirk: It's not in this package.

MapboxNavigation/NavigationMapView.swift
michaelkirk: It's so that we don't accidentally break the puck animation again, like what I fixed in #53.

I tried to explain the reasoning in this issue description — the method we were overriding was a private method. I patched it up in #53, but because it's still a private method it could break again. #53 has some more details which might be helpful for context.

Patrick-Kladek: Got it, thanks for linking the PR, helps a lot to understand.

hactar (approved):

PR #110, opened 2025-06-10
Missing frame error fix

IlyesDjari:
### Checklist

- [ ] I added an entry to `CHANGELOG.md` if knowledge of this change could be valuable to users.
- [ ] I linked any relevant issues from https://github.com/maplibre/maplibre-navigation-ios/issues

### Description
This pull request resolves a build failure that occurs when compiling the library with Xcode 16.2.

The Swift compiler in Xcode 16.2 now correctly flags an error in the CarPlayMapViewController.swift file. The loadView() method attempts to initialize NavigationMapView using NavigationMapView(), but the class does not provide an initializer that takes zero arguments.

This results in a 'Missing argument for parameter 'frame' in call' compile-time error, which prevents the library from being successfully built and used in projects targeting the latest iOS SDKs.

This PR fixes the build error by providing a valid frame during initialization, changing the call to NavigationMapView(frame: UIScreen.main.bounds). This change aligns with standard UIKit practices and allows the project to compile on Xcode 16.2 and newer.

### Infos for Reviewer
How to Reproduce: Attempt to build the library or an application that depends on it using Xcode 16.2. The build will fail with a compiler error in CarPlayMapViewController.swift.

Testing: This fix has been validated by forking the repository, applying the change, and integrating the fork into a React Native Expo application. After applying the patch, the project compiles, builds, and runs successfully on Xcode 16.2.
Importance: This change is essential for ensuring the library remains compatible with modern iOS development tools and can continue to be used by developers on the latest versions of Xcode.

Scope: This is a minimal, one-line fix that corrects a direct violation of the NavigationMapView's initializer contract. It has no known side effects and only affects the CarPlay-specific code path.


MaartenZonneveld:
@IlyesDjari The PR has 300+ files changed. That can't be right 😉

hactar:
> @IlyesDjari The PR has 300+ files changed. That can't be right 😉

I think @IlyesDjari didn't create a branch specific for his PR, and so commited further changes later. The original scope of this PR was https://github.com/maplibre/maplibre-navigation-ios/pull/110/commits/0775bb75878d455447967ba44aeb2ee0a316f583 if I remember correctly.

MaartenZonneveld:
Fixed in #113

PR #120, opened 2026-04-08
Update MapLibre logo in README, mention Ferrostar

louwers:
As we discussed during the MapLibre Navigation meeting today, we want to make people aware of the existence of Ferrostar, since it's easier to get it up and running with existing route planning software.

If (other) maintainers feel it's more appropriate as an "Alternatives" section at the bottom, please speak up!

cc @hactar


hactar (approved):
Sounds good. I agree that this should rather be at the top of the file. the iOS project still works fine, but there are "easier" solutions for more modern architectures now that are open source, and I think it would be good for developers to be aware of these solutions.

CommanderStorm (approved):

